PR #181, opened 2017-09-18
Rewrite MOC southern boundary extractor

xylar:
This merge fixes issues with the southern  boundary transect extracted for MOC regions through a substantial rewrite.  One significant change is that the edge sign is determined by the order of cells on an edge, not vertices (the latter of which may have caused issues with the previous approach).  The logic has also been simplified.


xylar:
I'm going to close this for now.  We can revisit when we have time.

xylar:
@milenaveneziani, should we revisit this.

milenaveneziani:
@xylar: wasn't this superseded by #147?

xylar:
No, I believe this came after #147 as a fix.  I believe we simply never were confident that the results were correct (but we also weren't sure they were not).  We needed to track down whether the differences we saw were expected because of the change in position of the southern transect (wrong currently, "right" here).

milenaveneziani:
I see, thanks for the reminder. Indeed, we have an open issue open because of this.

I'd say we can give it a try: maybe spend a couple of days max on it and see if we can make progress? How does this week look for you? I could also do early next week, if that works better for you.

xylar:
I can't get this PR to update properly.  I commit to the branch (on my old fork) but GitHub isn't noticing the changes here in the PR.  I think that's okay but it means the merge will need to happen at the command line, not here.

@milenaveneziani, this is ready for you to check out and review.  Note that the branch is on:
https://github.com/xylar/Old-MPAS-Tools/tree/fix_moc_southern_boundary_extractor
so you may need to check out my old remote.  I can push to my new remote instead and we can open a new PR if that gives you trouble.

milenaveneziani:
@xylar: I was able to checkout the branch on your old fork. Here are the commits I see:
```
938dfc1 (HEAD -> xylar/fix_moc_southern_boundary_extractor, xylar/Old-MPAS-Tools/fix_moc_southern_boundary_extractor) Add MOC southern transect script to package
67977b0 Rewrite MOC southern boundar extractor
```

xylar:
Great, that looks right!

milenaveneziani:
Hi @xylar. I created new transect files for 3 meshes (EC60to30, RRS18to6 and ARRM60to10).  Here are the things I notice: 1) the first 2 transects are identical. Even though the transect name of the first transect is Atlantic_MOC, in reality it's a repetition of the IndoPacific_MOC transect. The other 2 transects are fine. 2) A minor thing, but the variable `transectNames` has dimension `nRegions` instead of `nTransects`.

I was also thinking about how to test these transect files with MPAS-Analysis. Since now the transect name is Atlantic_MOC instead of simply Atlantic, `streamfunctionMOC` will fail. Would you suggest for me to make a temporary change in my local MPAS-Analysis repo to handle the new Atlantic_MOC name? [I have already locally changed the transect name to match what MPAS-A is expecting]

xylar:
@milenaveneziani, I just did a test and I'm able to reproduce the second of these issues but not first. I agree that `transectNames` has the wrong dimension name.  That should be easy to fix.  However, the `Atlantic_MOC` transect (index 0) is the correct transect for me:
atlantic_transect
All four transects are different and are the correct ones corresponding to each region name.

Could it be that your geojson file has the wrong region for the `Atlantic_MOC` somehow?  I'm not sure how that would happen, especially with the right name.

xylar:
> Would you suggest for me to make a temporary change in my local MPAS-Analysis repo to handle the new Atlantic_MOC name?

Hmm, there are a couple of options here.  I guess it would be easiest to just rename the features to what MPAS-Analysis expects.  This should be a trivial change to the script taken form geometric_features.  That way, we don't have to do extra work to be backwards compatible.

milenaveneziani:
> we don't have to do extra work to be backwards compatible

good point. We can just rename the transectNames, differently from the regionNames. I *think* MPAS-A will then be OK.

xylar:
> good point. We can just rename the transectNames, differently from the regionNames. I think MPAS-A will then be OK.

Can't the region names also be `Atlantic`, `Pacific`, etc. or does that mess up the online analysis members?

milenaveneziani:
Not sure, I'd have to check.
It could also be confusing, since we have the Atlantic_Basin region, for example.

milenaveneziani:
I checked the online am's: they look for regionGroupName, not for a specific name, so we should be fine there if we change the region and transect names.

xylar:
The issue in @milenaveneziani's tests above ended up being the use of the `--positive_lon` flag on the mask creator.  I just added better documentation suggesting that this flag should only be used in very rare circumstances of having a non-standard geojson file with positive longitudes.  None of the geometric_features files require this and many will not work correctly with it (because they cross the prime meridian).

milenaveneziani:
OK, so here are a few results I have so far.
What I did is create new southern transects for 3 meshes: EC60to30v3, RRS18to6v3 and ARRM60to10. Since I have ready available output from a EC60to30 run and a ARRM60to10 simulation on NERSC, I decided to use that and the MPAS-Analysis post-processing streamfunctionMOC script to compute the Atlantic MOC with the old and new transects and compare results.
I also let the script print out the net transport across the southern section for the Atlantic (the one corresponding with the southern transect), so that I can see how the net transports compare there.

Here is what I have for the **EC60to30** run.
Old transect:

mocTimeseries_EC60to30old

New transect:

mocTimeseries_EC60to30new

Net transport at southern section comparison:

TransportEC60to30

milenaveneziani:
And here is what I have for the **ARRM60to10** run.
Old transect:

mocTimeseries_ARRM60to10old

New transect:

mocTimeseries_ARRM60to10new

Net transport at southern section comparison:

TransportARRM60to10

milenaveneziani:
So, as you can see, we already start with consistently lower transports at the southern section with the new transects. I don't understand what would cause that. When we get to 26.5N we are between 0.5 and 1 Sv lower, depending on specific years.

Again, the issue here is not much about new vs old (since we don't know the 'truth'), but rather that we convince ourselves that the postprocessing script is doing what it is supposed to do or whether we still have some issue with the new transects (although signs seem correct, and there are no openings in the new transect to justify the different transport at the southern section).

Pinging @maltrud  and @mark-petersen as well for suggestions/input.

milenaveneziani:
Short update: I have looked carefully at both the offline calculation of transport across the southern transect and its online calculation, and they look basically identical (or as identical as a python code can be to a fortran code..). The only difference is that in the online code, the `transectEdgeMask` is applied to the global v*dx*dz, while in the offline code we use the `transectEdgeGlobalIDs` to identify the transect from the get go and compute the transport for those transect cells only. Just to be sure, I applied the global online method to the offline code, and results did not change up to the 15th decimal digit, so we are safe there.

In the process, I also found out that the new MOC transect maskfiles would *not* work with the MOC online code as is now, because the MOC am is looking for `transectsInGroup` and similar variables, not just `regionsInGroup` variables as we have in the new transect files.

This does not answer the question as to why the transport across the old and new transects are so different (and, more importantly, consistently shifted over a whole a time series), but at least gives me more confidence on the codes that we currently have (online and offline).

xylar:
> In the process, I also found out that the new MOC transect maskfiles would not work with the MOC online code as is now, because the MOC am is looking for transectsInGroup and similar variables, not just regionsInGroup variables as we have in the new transect files.

I believe this just means I need to copy `regionsInGroup` to `transectsInGroup`, correct?

milenaveneziani:
We can discuss this. I personally do not like having both, since regions are intrinsically related to transects. And the AM would benefit from a simplification related to having only one group.

xylar:
It's much easier to fix the python code than update the Fortran.  Plus the naming is a side effect of generality where transects can be completely independent of regions (outside the MOC context).  I don't think we want to break that generality.

milenaveneziani:
Yes, I think we do, for the MOC case. Regions and transects are just the same in that case.
But again, I think we can discuss what to do with the online code.

xylar:
@milenaveneziani, I just updated the script.  It should now produce a region/transect mask file that is compatible with the online MOC.  if that's not the case, please let me know.

milenaveneziani:
great, thanks.
I made the new transect file for the EC60to30: we can test the online AM with it now.

milenaveneziani:
@xylar: @vanroekel  noticed that the new transect file has the dimension `StrLen` equal to 1024 instead of the old 64. Is it the new `MaskCreator` tool that does this or is it the southern transect extractor?
I will try changing the dimension size for now and see if this solves the problem of the MOC AM.

xylar:
@milenaveneziani, the bigger number is likely necessary to handle the metadata that is included in the files.  I would think the better approach would be to modify the MOC AM to handle the longer string length.

milenaveneziani:
I agree that it would be nice to leave 1024, but, as you mentioned in the MPAS-Model PR, we could temporary change it to 64 and raise an issue in MPAS-Model to change this in the framework.

xylar:
I believe the remaining issue with 1024-character strings (instead of the expected 64) should be addressed by https://github.com/MPAS-Dev/MPAS-Tools/pull/275, followed by a release of MPAS-Tools and an update of the various COMPASS conda environments.

xylar:
@milenaveneziani, can we go ahead and merge this very old PR?  Presumably any remaining issues can be addressed later.

PR #219, opened 2018-05-24
modify bedTopography when we convert thickness

tongzhangice:
Let the script automatically change bedTopography if we want to use Albany's new thickness field. Plus, I now explicitly let beta smoothing use the 'min' method and stiffness the 'idw' method. If we want to deal with thickness, there will be no extrapolation, but still need to set smoothing iteration number to 0 to disable smoothing.

### DIFF
--- a/landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
+++ b/landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
@@ -35,19 +35,12 @@ mpas_exodus_var_dic = {"beta":"basal_friction", "thickness":"ice_thickness", "st
 # A mapping between mpas and exodus file. Add one if you need to manipulate a different variable!
 ice_density = 910.0
 ocean_density = 1028.0
-useKDTree = False
-# useKDTree is to find the surrounding nearest cells. Thus it only works for uniform meshes. But it's still useful for testing
-# Normally, always use the cellsOnCell method. If you do want to use this method, set useKDTree to True 
 
 exo_var_name = mpas_exodus_var_dic[options.var_name]
 
 dataset = Dataset(options.nc_file, 'r+')
 x = dataset.variables['xCell'][:]
 y = dataset.variables['yCell'][:]
-
-if useKDTree:
-    xy = np.array([x,y])
-    tree = spt.cKDTree(xy.T)
 
 exo = exodus(options.exo_file)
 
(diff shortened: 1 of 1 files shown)


xylar:
What's the status of this PR?  It looks like it needs to be rebased because of conflicts. Is it a work in progress?  Or Abandoned?

grid_gen/landice_grid_tools/conversion_exodus_init_to_mpasli_mesh.py
```diff
@@ -91,22 +93,35 @@
...
+        elif options.var_name == "thickness":
+            dataset.variables[options.var_name][0,index] = data_exo_layer[i]*1000.0
         else:
-            dataset.variables[options.var_name][0,index] = (data_exo_layer[i])
+            dataset.variables[options.var_name][0,index] = data_exo_layer[i]
 
         # The beta unit of the mpas mesh is Pa yr/m, not SI. 
         # This method may fail at the point where x or y = 0, while x_exo or y_exo is not
 
 elif (options.conversion_method == 'id'):
```
matthewhoffman: @tongzhangice , I am finally looking over this PR.  In looking again at this first part, I'm concerned that we have duplicated code for copying the fields depending on which conversion method is used.  As we are adding more and more fields, it seems like there will be challenge in maintaining both versions.  Maybe we should remove the coordinate method from the script.  What do you think?

grid_gen/landice_grid_tools/conversion_exodus_init_to_mpasli_mesh.py
```diff
@@ -141,69 +156,70 @@
...
+    while np.count_nonzero(keepCellMask) != nCells:
+        
+        keepCellMask = np.copy(keepCellMaskNew)
+        searchCells = np.where(keepCellMask==0)[0]
 
-while np.count_nonzero(keepCellMask) != nCells:
-    
-    keepCellMask = np.copy(keepCellMaskNew)
-    searchCells = np.where(keepCellMask==0)[0]
+        if useKDTree:
```
matthewhoffman: @tongzhangice , another question: In these parts of the script we also have code duplication between whether the kdTree or cellsOnCell methods are being used.  We could simplify the script if we remove the kdTree method and only leave the cellsOnCell method.  What do you think?

grid_gen/landice_grid_tools/conversion_exodus_init_to_mpasli_mesh.py
tongzhangice: @matthewhoffman personally I hope to keep the coordinate method, which is still useful when we do not have (or can not find) the global id file. But I do not have strong preference for it. I am OK with either way.

grid_gen/landice_grid_tools/conversion_exodus_init_to_mpasli_mesh.py
tongzhangice: @matthewhoffman I agree to remove it since we barely use the kdTree method anyway...

grid_gen/landice_grid_tools/conversion_exodus_init_to_mpasli_mesh.py
matthewhoffman: @tongzhangice ,  Ok that makes sense.  I think I can imagine a way to easily adjust the code so that  would remove the code duplication.  I will try that and let you have a look at it before I complete the merge.

matthewhoffman:
@tongzhangice , I rebased this PR to deal with some conflicts, and then added two commits to make the changes I suggested in my comments above.  I don't have write permission to your fork, so I can't push the changes to this PR, but I pushed them to my fork here:
https://github.com/matthewhoffman/MPAS-Tools/commits/add_exodus_conversion

Can you look it over and make sure you are ok with the changes?  If so, see if you can push this to your fork so that the PR updates.  Then I will merge it.  (I already checked to make sure my changes do not change the answers.)

tongzhangice:
> @tongzhangice , I rebased this PR to deal with some conflicts, and then added two commits to make the changes I suggested in my comments above. I don't have write permission to your fork, so I can't push the changes to this PR, but I pushed them to my fork here:
> https://github.com/matthewhoffman/MPAS-Tools/commits/add_exodus_conversion
> 
> Can you look it over and make sure you are ok with the changes? If so, see if you can push this to your fork so that the PR updates. Then I will merge it. (I already checked to make sure my changes do not change the answers.)

@matthewhoffman Hi Matt, I pushed your version to my branch. Can you take a look? I agree with your changes. I think it is now cleaner.

matthewhoffman:
@tongzhangice , thanks for looking it over.  For some reason I don't understand, the PR did not update correctly.  I see a different number of commits on the PR page (here) than on the page for the branch (https://github.com/tongzhangice/MPAS-Tools/commits/add_exodus_conversion).  Also, from what I can tell, you cherry-picked and merged my new commits.  However it will be easier for me to merge the version I had made that is rebased on master because it has no conflicts.  The way to do this is a force push my branch to your remote like the following command, assuming your remote is named 'tong' and mine is named 'matt':
```
git push tong +matt/add_exodus_conversion:add_exodus_conversion
```
(The + symbol allows the push command to overwrite commits already there.)  If you would rather not do that or are having a problem with it, send me an email and we can get together today and look at it.

PR #221, opened 2018-06-09
Directory reorganization

mark-petersen:
Add 'legacy' tag, remove defunct tools, then reorganize the directory structure of MPAS-Tools as follows:
- mesh_tools
- output_processing
- source_code_processing
- visualization
- atmosphere
- landice
- ocean
- seaice

where the first four include general mpas tools, and the last four are core specific.  We no longer organize by the language, i.e. python, matlab etc.  A separate commit was made for each delete and move, so they could be added with a single revert command later, if desired.


xylar:
> Yes, we discussed that again a few times and decided to keep them separate. We said that tools that are closely tied to things that get revised a lot should be in MPAS-Model - for example, COMPASS. But these tools, like mesh generation, vtk viewers, etc, are based on the MPAS mesh spec, which has been stable for a long time. We decided that it was better to have two separate repos for clarity, rather than make MPAS-Model serve all purposes. It allows us to be a bit looser on allowing more scripts to be added to MPAS-Tools.

Okay, that all sounds reasonable.  Let's make sure to keep any tools that are COMPASS specific out of this repo and in MPAS-Model, then.  I think we're doing that already.

PR #222, opened 2018-07-13
Ocean performance testing

divya-jaganathan:
Add scripts that generate performance curve plots and data for MPAS-Ocean.


ocean/performance_testing/performance_testing.py
```diff
@@ -30,9 +30,9 @@
     m1 = re.search("config_run_duration", line)
     if m1:
         parts = line.split("=", 1)[1]
-        timeparts = re.split(':|\'|_', parts)
-        simulated_time_in_sec = int(timeparts[1]) * 24 + int(
-            timeparts[2]) * 60 * 60 + int(timeparts[3]) * 60 + int(timeparts[4])
+        timeparts = re.split(':|\'|_|-', parts)
+        simulated_time_in_sec = int(timeparts[3]) * 24*60*60 + int(
+            timeparts[4]) * 60 * 60 + int(timeparts[5]) * 60 + int(timeparts[6])
```
mark-petersen: @divya-jaganathan thank you for your attention to detail here.  Computing time up to days is sufficient for a performance test script.

mark-petersen:
@divya-jaganathan Thank you for your work on these scripts.  I've been testing it, and it works beautifully.  I only altered some flag names and details on the plots.  Here is an example from wolf:
ec60to30_wolf_256_20180905_104312

mark-petersen:
@philipwjones I think this script will be extremely useful, for both ocean model developers and performance specialists.  On any machine, you can grab tarred run directories here:
https://zenodo.org/record/1252437#.W5FIppNKjUI
add a link to 
- metis
- ocean_model executable
- call_to_performance_testing.py (here)
- performance_testing.py (here)

On any machine log-in node, all you need is:
```
./call_to_performance_testing.py
```
This will submit a single job to the queue, and produce the default test of 64 through 2 by powers of 2, and auto-detect your machine.  Load modules before calling this script, and submission will keep the same modules.

Or, one can specify everything with flags.  This tests 128 to 16 cores by powers of two.
```
./call_to_performance_testing.py -M 128 -m 16 -r EC60to30
```

On cori, you have to specify cori-knl or cori-haswell, as follows:
```
./call_to_performance_testing.py -M 128 -m 16 -r EC60to30 -n cori-knl
```

mark-petersen:
@philipwjones Could you give this a try in the next week?  It would be good to hear any comments you might have.

mark-petersen:
@matthewhoffman, @akturner, @darincomeau, @mgduda  I just wanted to ping you on this.  It could be very useful for your own development.  I think the only core-specific thing is the name of the executable and the timer name.

philipwjones:
Tried this a few times on Cori-knl for a QU240 case and failed - first few failures were my fault, but when I finally got my act together, it appears to time out. So maybe needs some additional args to manage time limits or number of samples. 

Seems to be a useful quick-look tool for pure MPI runs on simple clusters. May need further work for a system like Summit where defining resource sets in slurm is more complex and requires more args.

mark-petersen:
Thanks for testing, @philipwjones.  It probably timed out because those cases have output and restart file writing on.  I plan to create a packaged test with all i/o off for exactly this type of testing, but otherwise it is easy to turn off in the streams file.

divya-jaganathan:
Hello Mark,

I hope you are doing well. Classes have started and I am settling down
slowly.

Apologies for the late response. I hadn't noticed the mail in a long time.
My Brown univ e-mail isn't working anymore, and this one got pushed to a
different folder. Thanks for updating me with how the script works, and for
improving it for friendlier usage. Please do let me know if I can do
anything more along similar lines.

And sorry again for this late a reply.

Best regards,
Divya

divya-jaganathan:
Hi Mark,

I can be reached over at the following e-mail addresses-
divya.jaganathan@icts.res.in
divjag2005@gmail.com

Thanks

Regards,
Divya

PR #224, opened 2018-08-08
Allows paraview vtk extractor extraction of single dimension time coord

pwolfram:
`xtime` is typically a 2D string.
`timeMonthly_avg_daysSinceStartOfSim` in contrast
is a float corresponding to days of simulation.

This PR allows extraction of fields with
`timeMonthly_avg_daysSinceStartOfSim` within
`paraview_vtk_field_extractor.py`.

### DIFF
--- a/visualization/paraview_vtk_field_extractor/utils.py
+++ b/visualization/paraview_vtk_field_extractor/utils.py
@@ -96,9 +96,13 @@ def setup_time_indices(fn_pattern, xtimeName):  # {{{
                 raise ValueError("xtime variable name {} not found in "
                                  "{}".format(xtimeName, file_name))
             local_times = []
-            xtime = nc_file.variables[xtimeName][:, :]
-            for index in range(xtime.shape[0]):
-                local_times.append(''.join(xtime[index, :]))
+            xtime = nc_file.variables[xtimeName]
+            if len(xtime.shape) == 2:
+                xtime = xtime[:, :]
+                for index in range(xtime.shape[0]):
+                    local_times.append(''.join(xtime[index, :]))
+            else:
+                local_times = xtime[:]
 
             if(len(local_times) == 0):
                 local_times = ['0']
--- a/visualization/paraview_vtk_field_extractor/utils.py
+++ b/visualization/paraview_vtk_field_extractor/utils.py
@@ -615,7 +615,7 @@ def build_cell_geom_lists(nc_file, output_32bit, lonlat):  # {{{
                                                          validVertices,
                                                          lonCell)
 
-    if nc_file.is_periodic == 'YES':
+    if nc_file.on_a_sphere == 'NO' and nc_file.is_periodic == 'YES':
         if lonlat:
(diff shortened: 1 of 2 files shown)


pwolfram:
@xylar, the primary use case here is to view data from monthly stat files, e.g., `mpaso.hist.am.timeSeriesStatsMonthly*nc`, via selection of the time dimension via something like `-x timeMonthly_avg_daysSinceStartOfSim` passed as an argument to `paraview_vtk_field_extractor.py`.

visualization/paraview_vtk_field_extractor/utils.py
```diff
@@ -96,9 +96,13 @@ def setup_time_indices(fn_pattern, xtimeName):  # {{{
                 raise ValueError("xtime variable name {} not found in "
                                  "{}".format(xtimeName, file_name))
             local_times = []
-            xtime = nc_file.variables[xtimeName][:, :]
-            for index in range(xtime.shape[0]):
-                local_times.append(''.join(xtime[index, :]))
+            xtime = nc_file.variables[xtimeName]
+            if len(xtime.shape) == 2:
+                xtime =xtime[:, :]
```
xylar: please fix the formatting around the equals

pwolfram:
@xylar, I think there are cases where having a non-string representation makes sense.  For example, `timeMonthly_avg_daysSinceStartOfSim` is a float, not the complex time string.  This PR allows numbers to be used as the time index, which is broadly helpful if just one piece of time information is available.  For most applications, `--xtime xtime_startMonthly` is better.  I just couldn't go this route with the data file I had this time.  

I've fixed the issue as you requested, thanks!

xylar:
@pwolfram, was this some old data file where `xtime_startMonthly` was not being written out?  All cases I'm aware of should now include `xtime` or `xtime_startMonthly` (or similar).

I agree that there's no harm in supporting a variable with a float time.

pwolfram:
@xylar, it is for a non-standard workflow so I'm not aware of this being a common workflow "bug" that needs correction outside this repo.

xylar:
@pwolfram, in the process of testing this PR, I found that there are three lines:
```
if nc_file.is_periodic == 'YES':
...
```
that need to be:
```
if nc_file.on_a_sphere == 'NO' and nc_file.is_periodic == 'YES':
...
```
This is because we don't store the attribute `is_periodic` for meshes that are on the sphere.  Would you be okay with me pushing a commit to this branch that fixes this issue (even though it is unrelated to this PR)?  If not, I can make a separate PR.

pwolfram:
@xylar, I'm ok with that because we aren't worried about bullet-proof regression testing / bug finding for this code and for this repo more generally.

pwolfram:
@xylar, I'll wait for your commit and then merge once you are ready.

xylar:
I'd recommend doing one last test of your own and then merging if all works as expected.

pwolfram:
@xylar, this is great -- thank you!

PR #225, opened 2018-08-24
Fixes lat-lon coordinates for MPAS-O output

pwolfram:
Fixes issue where non-used (effectively masked) verticesOnCell values
differ between 0 and nVertices, which is produced by MPAS-O.
The existing script assumes these values must always be 0 and this
commit ensures that assumption.

### DIFF
--- a/visualization/paraview_vtk_field_extractor/utils.py
+++ b/visualization/paraview_vtk_field_extractor/utils.py
@@ -604,6 +604,9 @@ def build_cell_geom_lists(nc_file, output_32bit, lonlat):  # {{{
 
     nEdgesOnCell = nc_file.variables['nEdgesOnCell'][:]
     verticesOnCell = nc_file.variables['verticesOnCell'][:, :] - 1
+    # MPAS-O sets non-masked values to total number of vertices instead of 0
+    # (as produced in mesh workflow)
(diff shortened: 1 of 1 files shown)


pwolfram:
# Testing
Tested with 
`~/Documents/MPAS_development/MPAS-Tools/visualization/paraview_vtk_field_extractor/paraview_vtk_field_extractor.py  -f ocean.nc  -v bottomDepth -o tmp --ignore_time -l`
using data at https://drive.google.com/open?id=1SCes0Uy4IPpHcmXYY4OqaMIBDGVhlbCZ

xylar:
@pwolfram, I think this is a perfectly reasonable fix.  I'll note that MPAS output files write out 0, rather than nVertices+1, for invalid vertices.  After subtracting 1, this means the python vertex index would be -1,  which ends up pointing to the last vertex.  This is why we hadn't seen this issue earlier.

Because we later mask out invalid vertices, it is only important that lines like:
https://github.com/MPAS-Dev/MPAS-Tools/blob/8d2ecd31e46dc5990059ba4951e8292446616a9f/visualization/paraview_vtk_field_extractor/utils.py#L984
can compute a value successfully.  Thus, it should be fine if invalid vertices are given any value in the range of 0 to nVertices-1, and 0 seems like the obvious choice.

PR #226, opened 2018-10-18
Fixes grid_rotate overwrite bug

MiCurry:
When the same name was given for the source and destination file, grid_rotate
would overwrite the source file with the rotation, instead of failing.

This fix allows grid_rotate to exit when the source filename is the same as the
destination filename as well as any other cp error.

### DIFF
--- a/mesh_tools/grid_rotate/grid_rotate.f90
+++ b/mesh_tools/grid_rotate/grid_rotate.f90
@@ -83,6 +83,7 @@ contains
       real (kind=RKIND) :: cx, cy, cz
 
       character(220) :: copyCmd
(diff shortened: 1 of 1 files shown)


MiCurry:
This PR resolves #100

mgduda:
@mark-petersen @matthewhoffman @akturner Do any of you use the `grid_rotate` tool in your workflows? If so, I could add you as reviewers; otherwise, I'll review this PR and merge after that.

mark-petersen:
We have used it in the past.  I don't think we use it in our current workflows. You should go ahead, those changes look very straightforward.

PR #228, opened 2018-11-01
Script to create coupling files for E3SM

mark-petersen:
Intended to begin process outlined here:
https://acme-climate.atlassian.net/wiki/spaces/OCNICE/pages/763854989/Simplify+E3SM+Ocean+File+Generation

to standardize the process of running new MPAS meshes in E3SM.


mark-petersen:
@xylar could you give this a first look?  I'm trying to learn good python form at the same time, so all detailed comments are welcome.  I did everything we talked about yesterday except test in python3.

mark-petersen:
Download test mesh and graph file here:
https://drive.google.com/open?id=1F9xtOipsehrZxW7odyFSz2o1dLlc7qCx

xylar:
This is great!  My only comment so far is, could you create a file `ocean/coupling_files_E3SM/readme.md` where you can say a little about what the script does and, importantly, document what its dependencies are (currently NCO and gpmetis) and how to install them with conda?  This doesn't have to be right away, but I think should be part of the process down the road.

xylar:
I tested the script on my laptop.  Here were my steps:
``` bash
conda create -y -n coupling_files_E3SM_py2.7 python=2.7 numpy configparser nco metis six
conda activate coupling_files_E3SM_py2.7
# change path to gpmetis
vim create_E3SM_MPAS_coupling_files.ini
./create_E3SM_MPAS_coupling_files.py
```
Everything ran as expected and the output looks reasonable as far as I can tell:
```
$ ls * */*
command_history
create_E3SM_MPAS_coupling_files.ini
create_E3SM_MPAS_coupling_files.py
graph.info
graph_partition_ocean/command_history
graph_partition_ocean/gpmetis
graph_partition_ocean/log.out
graph_partition_ocean/mpas-o.graph.info.181031
graph_partition_ocean/mpas-o.graph.info.181031.part.16
graph_partition_ocean/mpas-o.graph.info.181031.part.24
graph_partition_ocean/mpas-o.graph.info.181031.part.32
graph_partition_ocean/mpas-o.graph.info.181031.part.36
graph_partition_ocean/mpas-o.graph.info.181031.part.48
graph_partition_ocean/mpas-o.graph.info.181031.part.60
graph_partition_ocean/mpas-o.graph.info.181031.part.64
initial_condition_ocean/command_history
initial_condition_ocean/log.out
initial_condition_ocean/oUSDEQU300cr3.nc
initial_condition_ocean/oUSDEQU300cr3.nc_no_xtime.nc
oUSDEQU300cr3.nc
```

xylar:
Something you could do if you want to be a little fancy is you could see if gpmetis is in the path already (as it was for me) and not bother making a local link if it is.  The default in the config file could be just `gpmetis` without a path and that would indicate look in the path.  Since metis is a conda package, I don't see why it couldn't just be included in whatever environment was used to run the code, so maybe you could just assume gpmetis is in the path.

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.py
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env python
+"""
+This script creates coupling files needed to run an MPAS-Ocean and
+MPAS-seaice within E3SM.
+"""
+import os
+import shutil
+import subprocess
+import ConfigParser
```
xylar: To support python 3, this would need to change to:
```suggestion
ifrom six.moves.configparser import ConfigParser
```

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.py
```diff
@@ -0,0 +1,119 @@
...
+import numpy as np
+
+
+def main():
+
+    config = ConfigParser.ConfigParser()
```
xylar: To support python 3, this would become:
```suggestion
    config = ConfigParser()
```

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.py
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env python
+"""
+This script creates coupling files needed to run an MPAS-Ocean and
+MPAS-seaice within E3SM.
+"""
+import os
```
xylar: To support both python 2 and python 3, it  would be good to add (as the first import):
```suggestion
from __future__ import absolute_import, division, print_function, \
    unicode_literals
    
import os
```

xylar:
I tested with python 3. The three changes I noted above were required.  With those, the following worked great for me:
```
conda create -y -n coupling_files_E3SM_py3.6 python=3.6 numpy configparser nco metis six
conda activate coupling_files_E3SM_py3.6
# change path to gpmetis
vim create_E3SM_MPAS_coupling_files.ini
./create_E3SM_MPAS_coupling_files.py
```

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.py
xylar: By the way, `six` is a package with an overly cute name (2 x 3 = six) for supporting certain packages in python 2 and 3 at the same time where no other solution exists.

xylar:
Kristin, your new functions look good!  Maybe @mark-petersen can help with some code formatting (perhaps using `autopep8`).

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.py
```diff
@@ -0,0 +1,201 @@
...
+    input_file = config.get('main', 'input_file')
+
+    # create links
+    make_link('../' + input_file, input_file) 
+  
+    # link to MPAS Tools directory 
+    MPAS_Tools = config.get('main', 'MPAS_Tools')
+    make_link(MPAS_Tools + 
+    '/mesh_tools/create_SCRIP_files/create_SCRIP_file_from_MPAS_mesh.py', 
+    'create_SCRIP_file_from_MPAS_mesh.py')
```
xylar: This if fine for now but we might want to just call functions from python scripts down the road rather than calling a python script from another python script.  Just something to consider.

xylar: On second thought, it would probably be an unnecessarily confusing deviation from the workflow here to call functions in other python scripts.  This is good as it is.

xylar:
@mark-petersen, great!  I forgot to mention that the `.nc` inside the file name of the no-xtime file was confusing.  Nice to see a clean way to deal with that.

mark-petersen:
@xylar and @kristin-hoch, I'd like to dust this off and work on it again in the coming week.

If you want to test, it is easy to use the QU240 here:
https://zenodo.org/record/1252437#.XbbUrJNKjUI
or you can create new files with COMPASS.

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.py
```diff
@@ -0,0 +1,346 @@
...
+    print("****** " + function_name + " ******")
+
+    if config.get(function_name, 'enable') == False:
+        print("Disabled in .ini file")
+        return
+    make_dir(function_name)
+    os.chdir(function_name)
+
+    try:
+        globals()[function_name](function_name, config)
```
mark-petersen: @xylar could you look at this little function I added? I just wanted to put all the repeated code together in the same function, so this is really a wrapper. The function name is passed in. I got this `globals()` method from an on-line post, and it works. Do you have an opinion about that method?

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.py
```diff
@@ -0,0 +1,346 @@
...
+        '/usr/projects/climate/mpeterse/repos/tools/compiled_mesh_conversion_tools/mesh_tools/mesh_conversion_tools/bin/gr/gnu/MpasMaskCreator.x',
+        'MpasMaskCreator.x')
+    make_link('../' + mesh_name + '.nc', mesh_name + '.nc')
+    make_link(
+        '/usr/projects/climate/mpeterse/analysis_input_files/geojson_files/SingleRegionAtlanticWTransportTransects.geojson',
+        'SingleRegionAtlanticWTransportTransects.geojson')
+
+    argv = ['./MpasMaskCreator.x', mesh_name + '.nc', maskfile,
+            '-f', 'SingleRegionAtlanticWTransportTransects.geojson']
+    run_command(argv)
```
xylar: This should use the MpasMaskCreator.x from the conda environment rather than the local link, I think.

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.py
```diff
@@ -0,0 +1,346 @@
...
+    maskfile = 'masks_SingleRegionAtlanticWTransportTransects.' + mesh_name + '.nc'
+
+    # make links
+    make_link(
+        '/usr/projects/climate/mpeterse/repos/tools/compiled_mesh_conversion_tools/mesh_tools/mesh_conversion_tools/bin/gr/gnu/MpasMaskCreator.x',
+        'MpasMaskCreator.x')
+    make_link('../' + mesh_name + '.nc', mesh_name + '.nc')
+    make_link(
+        '/usr/projects/climate/mpeterse/analysis_input_files/geojson_files/SingleRegionAtlanticWTransportTransects.geojson',
+        'SingleRegionAtlanticWTransportTransects.geojson')
```
xylar: This should use `geometric_features` to pull down this feature rather than having a hard-coded path.  Better yet, we need to use the tools that @milenaveneziani and I have been using rather than this outdated, single, wrong transect.

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.ini
```diff
@@ -0,0 +1,41 @@
...
+# module use /usr/projects/climate/SHARED_CLIMATE/modulefiles/all
+# module load gcc/5.3.0
+# module load netcdf/4.4.1
+# module unload python
+# module load e3sm-unified
+
+[main]
+date_string = 181120
+mesh_name = oQU240
+MPAS_Tools = /usr/projects/climate/mpeterse/repos/tools/create_scrip_file_python3/
```
xylar: Any tools you need here should be added to the `mpas_tools` package rather than pointing to a path.

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.py
```diff
@@ -0,0 +1,346 @@
...
+    mesh_name = config.get('main', 'mesh_name')
+    date_string = config.get('main', 'date_string')
+    coupling_output = config.get('main', 'coupling_file_output')
+
+    maskfile = 'masks_SingleRegionAtlanticWTransportTransects.' + mesh_name + '.nc'
+
+    # make links
+    make_link(
+        '/usr/projects/climate/mpeterse/repos/tools/compiled_mesh_conversion_tools/mesh_tools/mesh_conversion_tools/bin/gr/gnu/MpasMaskCreator.x',
+        'MpasMaskCreator.x')
```
xylar: This path shouldn't be here.  you should use this tools from the `mpas_tools` package and assume it's in the path.

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.py
```diff
@@ -0,0 +1,346 @@
...
+
+def coupling_function(function_name,config):
+
+    print("****** " + function_name + " ******")
+
+    if config.get(function_name, 'enable') == False:
+        print("Disabled in .ini file")
+        return
+    make_dir(function_name)
+    os.chdir(function_name)
```
xylar: you need to store the name of the current directory before calling `chdir` and then change back to that directory after the function call.  This is because your `try`/`except` below could mean you don't end up back where you started if a function fails.

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.py
xylar: I've got to say that this is not how I would have done it but it looks workable to me.

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.py
```diff
@@ -0,0 +1,346 @@
...
+    config = ConfigParser()
+    config.read("create_E3SM_MPAS_coupling_files.ini")
+
+    coupling_function('initial_condition_ocean',config)
+    coupling_function('graph_partition_ocean',config)
+    coupling_function('initial_condition_seaice',config)
+    coupling_function('scrip',config)
+    coupling_function('transects_and_regions',config)
+    coupling_function('mapping_Gcase',config)
+    coupling_function('domain',config)
```
xylar: A more "pythonic" way to do this would be:
```
for function_name in ['initial_condition_ocean', 'graph_partition_ocean', ...]:
    coupling_function(function_name, config)
```
`coupling_function` is pretty small so its contents could just go inside of the for loop above.:
```
for function_name in ['initial_condition_ocean', 'graph_partition_ocean', ...]:
    print("****** " + function_name + " ******")

    if config.get(function_name, 'enable') == False:
        print("Disabled in .ini file")
        continue
    make_dir(function_name)
    os.chdir(function_name)
    ...
```

xylar:
@mark-petersen, could you run `autopep8` on this script?  I'd like to stick with that standard for any new scripts that go into the repo.

Also, we should add this script to the conda package as part of this PR.

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.py
```diff
@@ -0,0 +1,346 @@
...
+            mesh_name + '_no_xtime.nc'
+            ]
+    run_command(args)
+
+    # create link to output directory
+    make_dir('../' + coupling_output + '/ocn/mpas-o/' + mesh_name)
+    os.chdir('../' + coupling_output + '/ocn/mpas-o/' + mesh_name)
+    make_link('../../../../../' + function_name + '/' + mesh_name + '_no_xtime.nc',
+        mesh_name + '_no_xtime.nc')
+    os.chdir('../../../../..')
```
xylar: This and other such calls are not needed if you `chdir` back to the original directory outside of this function call, which is much safer.  See my comment above.

ocean/coupling_files_E3SM/create_E3SM_MPAS_coupling_files.py
mark-petersen: The [other method](https://www.programiz.com/python-programming/methods/built-in/eval) is to use:
```
eval(expression, globals=None, locals=None)
eval() Parameters
The eval() takes three parameters:

expression - this string as parsed and evaluated as a Python expression
globals (optional) - a dictionary
locals (optional)- a mapping object. Dictionary is the standard and commonly used mapping type in Python
```
I tried that but I had trouble with the dictionary so used `globals()` instead. Is `eval` what you were thinking of?

xylar: No, I think what you've done is likely safer than `eval`.  My sense is that you have to be careful about using `eval` especially if you include any user-supplied strings because it's a potential opportunity for malicious code to be added someplace like your config file.  I realize this isn't a concern for us but that's the reason I've got the sense of not using `eval` or doing so with great care.

I've never tried to call a function based on its name in a string before, to be honest.  I would have written this code in a way where that wasn't done and the function was just called as normal.  You can get a function's name with the `__name__` attribute so I would have done it the opposite way, like this:
```
for function in [initial_condition_ocean, graph_partition_ocean, ...]:
    function_name = function.__name__
    print("****** " + function_name + " ******")

    if config.get(function_name, 'enable') == False:
        print("Disabled in .ini file")
        continue
    make_dir(function_name)
    os.chdir(function_name)
    function(function_name, config)
    ...
```

PR #231, opened 2018-12-03
Add error handling for sphereRadius=0

### LINKED ISSUE #132: infinite values for "grid_area" when creating SCRIP file from land ice mesh with "sphere_radius" attribute = 0
@douglasjacobsen and @matthewhoffman, I've come across a problem with some recent updates to the "create_SCRIP_file_from_MPAS_mesh.py" script. 

The change on line 69 (line 70 in new code) here https://github.com/MPAS-Dev/MPAS-Tools/commit/ce11b3d1597ffbbbae844dd14f26563d47a081f3 causes division by zero with land ice meshes where the "sphere_radius" attribute is set to 0 (previously, sphere radius value was hard coded into this python script). In turn, this leads to inf values for the "grid_area" field in the resulting SCRIP file.

Should the land ice folks be altering their meshes so that this value is defined correctly even when using planar meshes?

stephenprice:
Add simple error handling to avoid division by zero when sphereRadius field is set to 0,
which is commonly the case for planar meshes. Setting to 1 in this case ensures that the
grid_area values in the output SCRIP file take on the same values as the areaCell field
in the input MPAS input mesh file.

closes #132

### DIFF
--- a/mesh_tools/create_SCRIP_files/create_SCRIP_file_from_MPAS_mesh.py
+++ b/mesh_tools/create_SCRIP_files/create_SCRIP_file_from_MPAS_mesh.py
@@ -53,6 +53,10 @@ maxVertices = len(fin.dimensions['maxEdges'])
 areaCell = fin.variables['areaCell'][:]
 sphereRadius = float(fin.sphere_radius)
 
+if sphereRadius <= 0:
+    print " -- field 'sphereRadius' = 0 (using planar mesh?); setting = 1 "
+    sphereRadius = 1.0
+
 if options.landiceMasks:
     landIceMask = fin.variables['landIceMask'][:]
 
--- a/mesh_tools/create_SCRIP_files/create_SCRIP_file_from_MPAS_mesh.py
+++ b/mesh_tools/create_SCRIP_files/create_SCRIP_file_from_MPAS_mesh.py
@@ -52,10 +52,14 @@ nCells = len(fin.dimensions['nCells'])
 maxVertices = len(fin.dimensions['maxEdges'])
 areaCell = fin.variables['areaCell'][:]
 sphereRadius = float(fin.sphere_radius)
+on_a_sphere = str(fin.on_a_sphere)
+
 
 if sphereRadius <= 0:
-    print " -- field 'sphereRadius' = 0 (using planar mesh?); setting = 1 "
-    sphereRadius = 1.0
+    print " -- WARNING: conservative remapping is NOT possible when 'sphereRadius' <= 0 because 'grid_area' field will be infinite (from division by 0)"
+
+if on_a_sphere == "NO":
+    print " -- WARNING: 'on_a_sphere' attribute is 'NO', which means that there may be some disagreement regarding area between the planar (source) and spherical (target) mesh"
 
 if options.landiceMasks:
     landIceMask = fin.variables['landIceMask'][:]


stephenprice:
@xylar , @matthewhoffman -- I tested this on a recent Ant. mesh file and it appears to work fine. It prints a warning to let you know that sphereRadius=0 just in case that is not supposed to be the case. I confirmed the values for the 'grid_area' (SCRIP) and 'areaCell' (MPAS) fields are identical, assuming that is supposed to be the case.

stephenprice:
Thanks @xylar. I'm happy to merge myself unless @matthewhoffman wants a look first.

xylar:
Actually, I looked a bit further and I'm concerned about this conversion.  If you are using the `areaCell` e.g. to do conservative remapping and SCRIP expects areas in square radians, the current conversion will instead leave the areas as square meters.  The conversion from areas on a polar stereographic projection to the corresponding area on the sphere is not trivial and probably deserves a separate script or a more complicated bit of logic with an extra flag.

I still think this fix is okay for now but you may not be getting proper mapping files from polar stereographic to the sphere with this code (regardless of this fix).

stephenprice:
I'll let it sit on my fork for the time being then.

xylar:
@stephenprice and @matthewhoffman, I've given this a bit more thought.  

The way the mapping file make use of areas (in conservative mode) is:
equation
where w is a weight and  equation is the fraction of the source cell i contributing to destination cell j.

I think we want area-integrated flux to be conserved, so that the models agree on flux*area for a given area even if they don't agree on either the flux or the area themselves.  If that's true, I think we would just use MALI areas as they are defined on the polar stereographic grid.  If MALI thinks a cell is smaller than it really is on the sphere, the flux that gets passed to the ocean or atmosphere will be slightly smaller than what came out of MALI to compensate for the larger area on the sphere.

If we agree on all of that, I think all you need to do is define a correct radius of the sphere (in meters) even for MALI planar grids so that areas get converted to square radians in the correct way.  Then, fluxes *should* be conservative in the sense that the flux integrated over what MALI thinks the total area is would come out to be the same as the remapped flux integrated over what the ocean or atmosphere thinks the area is.

stephenprice:
Thanks @xylar. I'll talk about this w/ @matthewhoffman some more tomorrow. 

I was also going to say that if this does need to sit for a bit before we figure out a (more) permanent solution, it would probably be a good idea to commit the more explicit error message for the case where sphereRadius=0 (i.e., print a very explicit warning that the conservative remapping won't work -- the current version of the code just gives a divide by zero warning but it's not immediately clear what the implications of that are since it still outputs the scrip file).

xylar:
Yes, I think that might be a good use for this pull request.  You could replace area_cell with NaNs if you wanted but I guess infinities should be fine, too.

Do you foresee any downside to giving your planar meshes a real-world earth radius?  Or do you think users will get confused and think it's not a polar stereographic mesh after all?

matthewhoffman:
@xylar, I agree with "I think we want area-integrated flux to be conserved, so that the models agree on flux*area for a given area even if they don't agree on either the flux or the area themselves."

I don't yet totally follow your proposal, though.  In your formula, do the individual component script files supply their own internally calculated areas, but then f is calculated by the regridder based on the lat/lon positions of the two sets of cell boundaries?  When you say "define a correct radius of the sphere" do you mean the sphere radius that the ocean or atmosphere is using?  Or the radius that is assumed by the polar steregraphic grid?

stephenprice:
@matthewhoffman -- @xylar and I talked about this a bit this morning. The radius he has in mind is that used by the other cores. He thinks f_ij is calculated as you mention above (based on lat/lon coords. of cell boundaries on the two meshes). 

We talked about doing the following:
1) modifying the landice core mesh files to have 'sphereRadius' consistent with ocean and sea ice
2) using the 'on_a_sphere' (="NO" for our polar stereo meshes) attribute as a way to flag the scrip file generation code to throw a warning about that areas between the spherical planar meshes will be slightly off
3) if there are still cases where 'sphereRadius=0' (e.g. periodic mesh), use that to throw a warning that conservative remapping cannot be used (because the field 'grid_area' will be infinite in that case)

I'm looking through the code right now trying to determine where/how 'sphereRadius' is specified for the diff. cores.

xylar:
Yep, what @stephenprice is exactly what I was thinking.  It would be important to verify that `f_ij` is computed independent of A_si and A_dj before we fully trust the proposed approach but I'm pretty confident that's the case (a simple test would be to try different nonsense values for A_si and A_dj) and to see what effect they have on the resulting mapping files, but it may also be possible to dig into the documentation of ESMF_RegridWeightGen and/or Tempest to see if this is explicitly stated somewhere.

stephenprice:
Is Charlie Z. likely to know more specifics on how F_ij is calculated?

matthewhoffman:
@xylar, @stephenprice and I are debating whether to 'fix' the sphere radius in the tool that creates a generic MPAS grid:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/mesh_tools/mesh_conversion_tools/mpas_mesh_converter.cpp#L2420
or in the land ice tool that creates a landice mesh from a generic MPAS grid here:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/mesh_tools/mesh_conversion_tools/mpas_mesh_converter.cpp#L2420

Either place will work, but the question is if we as an MPAS community (including atm) wants a hard-coded sphere radius in all planar meshes.  Thoughts?

cc: @mgduda

xylar:
@matthewhoffman, I think the second link should be to something else, right?  I'm guessing the latter would be more appropriate.

I don't think there is meant to be universal agreement across MPAS components about what the sphere radius is, so I would think we wouldn't want to put it in the mesh converter.  I also think that, aside from the specific case of planar meshes with real-world latitude and longitude (only MALI meshes at the moment) where you could conceivably making SCRIP files, we wouldn't care about the sphere radius and zero is a good indication that it's nonsense.

xylar:
We might want to alter:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/mesh_tools/mesh_conversion_tools/mpas_mesh_converter.cpp#L2420
to pass through the sphere radius unchanged if it exists, though, rather than setting it to zero.  Presumably, you would first create your MALI grid and then run it through the mesh converter and you probably wouldn't want the sphere radius set to zero and then have to recover it manually.

matthewhoffman:
@xylar , I agree with your suggestions.  I think it's ok if the mesh converter always sets sphere radius to 0 for planar meshes - our proposed change to the LI grid creation tool will always overwrite that with the hard-coded sphere radius value we choose.

xylar:
So `create_landice_grid_from_generic_MPAS_grid.py` would always be run *after* `MpasMeshConverter.x`?

stephenprice:
"So create_landice_grid_from_generic_MPAS_grid.py would always be run after MpasMeshConverter.x?"

Yes, that's correct.

xylar:
Yeah, then I think that's the best solution.  We can leave the mesh converter unchanged.

stephenprice:
Thanks @xylar. Unless I hear otherwise @matthewhoffman, then I'll proceed with making corrections to create_landice_grid_from_generic_MPAS_grid.py as we discussed earlier.

matthewhoffman:
@stephenprice , I think all we need to do is remove
" and getattr(filein, 'on_a_sphere') == "YES             "
 from this line:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/landice/landice_grid_tools/create_landice_grid_from_generic_MPAS_grid.py#L52
and then ensure the sphere radius on line 12 is the value we want.

xylar:
Sphere radius will come in a zero from periodic_hex:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/mesh_tools/periodic_hex/module_write_netcdf.F#L116
so you'll need to hard code it with some value in your script, I think.

For your reference, the radius of the earth is a framework-level constant in MPAS:
https://github.com/MPAS-Dev/MPAS-Model/blob/master/src/framework/mpas_constants.F#L26

stephenprice:
Thanks @xylar -- I was just trying to find out where radius of earth was set in MPAS. And yes, we expect sphereRadius will be 0 coming out of periodic_hex. We'll update it in our create_landice_grid etc. script.

xylar:
> Is Charlie Z. likely to know more specifics on how F_ij is calculated?

I missed this comment.  I would think he might.  @czender, do you know if the source and destination areas actually go into computing the fraction of overlap F_ij between a source cell i and destination cell j?  Or is this just determined based on lat/lon of the corners of the source and destination cells?  I think the latter is likely the case but we want to be sure.  Feel free to browse the discussion above or ask us questions if you need more context.

stephenprice:
@xylar -- please take another look at this when you have a chance. Now sphereRadius is no longer changed here but warnings are given based on its value and that of the 'on_a_sphere' attribute. Below is my test output:

**_Script screen output for MPAS-LI mesh w/ radius set to 0 and on_a_sphere = NO:_**

./create_SCRIP_file_from_MPAS_mesh.py -m temp.nc -s scrip.nc &
 **-- WARNING: conservative remapping is NOT possible when 'sphereRadius' <= 0 because 'grid_area' field will be infinite (from division by 0)
 -- WARNING: 'on_a_sphere' attribute is 'NO', which means that there may be some disagreement regarding area between the planar (source) and spherical (target) mesh**
./create_SCRIP_file_from_MPAS_mesh.py:90: RuntimeWarning: divide by zero encountered in divide
  grid_area[:] = areaCell[:]/ ( sphereRadius**2 ) # SCRIP uses square radians
Input latCell min/max values (radians): -1.56918998564 -1.05258320283
Input lonCell min/max values (radians): -3.13805064221 3.13800664244
Calculated grid_center_lat min/max values (radians): -1.56918998564 -1.05258320283
Calculated grid_center_lon min/max values (radians): -3.13805064221 3.13800664244
**Calculated grid_area min/max values (sq radians): inf inf**
Creation of SCRIP file is complete.

**_Script screen output for MPAS-LI mesh w/ radius set to 6371229.0 and on_a_sphere = NO:_**

./create_SCRIP_file_from_MPAS_mesh.py -m temp2.nc -s scrip2.nc &
 **-- WARNING: 'on_a_sphere' attribute is 'NO', which means that there may be some disagreement regarding area between the planar (source) and spherical (target) mesh**
Input latCell min/max values (radians): -1.56918998564 -1.05258320283
Input lonCell min/max values (radians): -3.13805064221 3.13800664244
Calculated grid_center_lat min/max values (radians): -1.56918998564 -1.05258320283
Calculated grid_center_lon min/max values (radians): -3.13805064221 3.13800664244
**Calculated grid_area min/max values (sq radians): 8.53383404003e-06 8.53383404003e-06**
Creation of SCRIP file is complete.

Assuming we fix the sphereRadius problem elsewhere, I think this does what we need in terms of improving the error handling of this script.

czender:
ESMF_RegridWeightGen computes the overlap fraction itself based on cell vertices. It does not use user-supplied areas unless specifically told to (with `--user_areas`). Hope this helps.

stephenprice:
@czender thanks! Xylar, sounds like you are ok w/ me merging this then?

PR #232, opened 2019-01-04
Add meshDensity field to doubly-periodic hexagonal grid.nc files

mgduda:
This merge adds a 'meshDensity' field, which is 1.0 everywhere, to the output of the doubly-periodic
mesh generation code to comply with the MPAS mesh spec v1.0.


pwolfram:
Awesome, thanks for doing this @mgduda and @matthewhoffman!

PR #236, opened 2019-01-16
Support xtime=none command-line argument

### LINKED ISSUE #235: Support for files with Time dimension but without xtime variable in VTK extractor
@bradyrx as requested a way to handle a file without an `xtime` or equivalent time variable.  Currently the VTK extractor can only pull out the first time slice of such a file with `--ignore_time`.

xylar:
With this value, Time is still extracted but without `xtime` (so that extracted time values are just time indices).

closes #235


xylar:
@bradyrx, let me know if this meets your needs.

xylar:
Okay, great!

PR #241, opened 2019-01-30
Remove shallow-water IC fields from periodic hex meshes

mgduda:
This merge removes shallow-water initial conditions from periodic hex meshes.

The periodic_hex mesh generator previously added fields to the grid.nc
file to serve as placeholders for ICs for the shallow-water model.
However, these fields are not part of the MPAS mesh specification, and
have been removed. Specifically, the following fields are deleted in
this commit:
```
  fEdge
  fVertex
  h_s
  u
  uBC
  v
  h
  vh
  circulation
  vorticity
  ke
  tracers
```
Additionally, the dimensions nVertLevels and nTracers have also been
removed.

### DIFF
--- a/mesh_tools/periodic_hex/module_write_netcdf.F
+++ b/mesh_tools/periodic_hex/module_write_netcdf.F
@@ -9,8 +9,6 @@ module write_netcdf
    integer :: wrDimIDmaxEdges2
    integer :: wrDimIDTWO
    integer :: wrDimIDvertexDegree
-   integer :: wrDimIDnVertLevels
-   integer :: wrDimIDnTracers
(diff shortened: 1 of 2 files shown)


mgduda:
I'm actually not sure which MPAS cores were using the `uBC` field that has been removed in this PR, but in case its absence would be a problem, we could perhaps leave it in the mesh file.

matthewhoffman:
@mgduda , this looks great to me.  The presence of those fields was always a minor annoyance to me.  I believe that all of our cores create new mesh files from the output of periodic_hex and don't actually use the meshes it generates directly.

PR #242, opened 2019-02-11
Add tool to extract ESM coupler history files to MPAS component grids

matthewhoffman:
This tool was written a couple years ago by Jeremy Fyke.
It takes a coupler history file as input, as well as optional MPAS
component files containing a grid specification for that component,
and outputs the coupler fields on the MPAS mesh.

### DIFF
--- a/visualization/e3sm_cpl_field_conversion/extract_mpas_cpl_fields.sh
+++ b/visualization/e3sm_cpl_field_conversion/extract_mpas_cpl_fields.sh
@@ -4,33 +4,24 @@ usage()
 {
 cat<<EOF
 usage: $0 options
-
 This script:
-
 1) extracts MPAS[LI,O,CICE]-generated/directed input/output coupler fields from coupler history files.
 2) converts the dimension naming to MPAS standards.
 3) appends all these fields, along with relevant mesh descriptor fields, to component-segregated output file(s) MPAS[LI,O,CICE]_CPL_fields.nc.
-
 mpas*_CPL_fields.nc can then be read by, e.g., ParaView.
-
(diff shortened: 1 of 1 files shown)


matthewhoffman:
@jonbob , @mark-petersen , @xylar , @darincomeau , @akturner, @JeremyFyke : This is a script Jeremy wrote a couple years ago that is useful for debugging/visualizing ESM coupler output.  I'm not sure where the best place to commit it is, but this was the most logical place I could come up with.  Feel free to suggest putting it elsewhere if you feel strongly about it.

@mgduda , this script currently only supports ocean, land ice, and sea ice, but it could also be applied for MPAS-A grids in CESM if you ever wanted to add that capability.

mark-petersen:
@matthewhoffman thanks for adding this.  The priority is that useful scripts get placed _somewhere_ on a repo. This seems like a fine place.   There is one other repo:
https://github.com/MPAS-Dev/MPAS-Quickviz
which was meant for unpolished viz/analysis scripts with no guarantees attached.  We could make a coupled directory there for this script, if you think it fits that category better.

matthewhoffman:
@mark-petersen , thanks for the feedback.  I think this is definitely a better place than Quickviz.  The other place I considered was somewhere in the E3SM repo/ecosystem, but I prefer to have it in MPAS-Tools.

xylar:
This seems like a fine place to store this tool.  I don't have any particular comments on what it does or anything like that.  Feel free to merge when you're ready.

darincomeau:
Glad this will be going here (somewhere) - I end up digging around my laptop anytime I want to use it.

PR #244, opened 2019-02-19
Add a script that performs vector reconstruction

xylar:
The script requires that `coeffs_reconstruct` has been stored in a NetCDF file.  This is most easily accomplished by adding a stream to the run (in froward mode):
```xml
<stream name="vector_reconstruction"
        clobber_mode="truncate"
        type="output"
        output_interval="0000-00-00_00:00:01"
        filename_template="vector_reconstruction.nc">

    <var name="coeffs_reconstruct"/>
</stream>
```
and running for one time step.


xylar:
@maltrud, this is supposed to help you with reconstructing zonal and meridional components of the bolus velocity, as you requested.

This script should run just fine in the e3sm-unified environment on whichever machine you're using.  Let me know if that gives you trouble.

You'll need to download [vector_reconstruction_oEC60to30v3.nc](https://drive.google.com/open?id=1ZodrZQC_byms31wYzYVDc9IkEagQf_bM) which I've put on my Google Drive for lack of anywhere else to store it permanently.  If this isn't just a one-off, it would be good to find a place to keep these kinds of things.

Usage is something like:
```bash
./vector_reconstruction.py -m mpaso.rst.0501-01-01_00000.nc -i mpaso.hist.am.timeSeriesStatsMonthly.0001-01-01.nc -w vector_reconstruction_oEC60to30v3.nc -v timeMonthly_avg_normalVelocity --outVariables velocity -o reconstructed.nc
```
The resulting NetCDF file looks like:
```
$ ncdump -h reconstructed.nc | more
netcdf reconstructed {
dimensions:
	Time = 1 ;
	nCells = 235160 ;
	nVertLevels = 60 ;
variables:
	double velocityX(Time, nCells, nVertLevels) ;
		velocityX:_FillValue = 9.96920996838687e+36 ;
	double velocityY(Time, nCells, nVertLevels) ;
		velocityY:_FillValue = 9.96920996838687e+36 ;
	double velocityZ(Time, nCells, nVertLevels) ;
		velocityZ:_FillValue = 9.96920996838687e+36 ;
...
```
You should be able to figure out what the flags are with `./vector_reconstruction.py --help` but let me know if it's not clear.

Most importantly, please let me know if you need different functionality.  For example, the script doesn't try to read in or write out multiple NetCDF files right now.  It doesn't try to append the new variables to an existing file.  It doesn't try to copy any attributes from the original field.  If some other behavior would be more convenient, let me know.

xylar:
## Testing

I verified that zonal and meridional velocities reconstructed with the tool look visually identical (in ParaView) to those computed in the model.  I didn't try to compute the difference between the two, but presumably they won't be identical due to order of operations.

xylar:
@mgduda, @akturner  and @matthewhoffman, I'm not sure if you have any interest in this but I wrote it to be general enough to work with any field where the MPAS vector reconstruction code is used.  

I made a framework directory and put this script directly there, but I'm open to other organization if you prefer.

framework/vector_reconstruction.py
```diff
@@ -0,0 +1,201 @@
...
+field, given the field on edge normals.
+
+This tool requires that the field 'coeffs_reconstruct' has been saved to a
+NetCDF file.  The simplest way to do this is to include the following stream
+in a forward run:
+
+<stream name="vector_reconstruction"
+        clobber_mode="truncate"
+        type="output"
+        output_interval="0000-00-00_00:00:01"
```
mgduda: If the coefficients don't change in time, I think one could also just use `output_interval="initial_only"`.

mgduda:
This looks really helpful! Would it make sense to name the new directory `operators/`, since the vector reconstruction code in the model is found in the `src/operators/` directory?

framework/vector_reconstruction.py
xylar:
```suggestion
        output_interval="initial_only"
```

xylar:
@mgduda, I'll rename the directory as you suggest.  I just incorporated the other change you suggested.  thanks very much!

xylar:
@maltrud, do you have time to try this out?

xylar:
@mark-petersen and @maltrud, could you test and/or approve this PR?

Issue #245, opened 2019-02-23
Creating a conda-forge recipe for mesh-conversion tools, planar_hex, etc.


xylar:
Having demonstrated in #243 that we could build the mesh conversion tools in a conda package, I would like to make a conda-forge recipe for those tools plus the basic mesh generation tools (periodic_quad and periodic_hex) so that COMPASS and other testing suites can be used more easily across a variety of machines and by a variety of users.

This would likely require pulling those tools out of this repo and putting them in their own, smaller repo that could have proper releases with tags and such.  Obviously, this would need to be discussed with the MPAS developer team(s) to make sure everyone is on board.

This issue is meant to be the start of that conversation.

cc @mark-petersen, @matthewhoffman, @akturner, @mgduda

xylar:
Looking at `periodic_hex`, I think it would be a 1-2 day project to make a python version.  It would probably take another few days at most to pull in all the capabilities from `planar_grid_transformations` and make a better python package of the whole thing.  I think that's probably the way to go.

Maybe the way forward with this whole idea of a separate `MPAS-Mesh-Tools` repo is for me to start working on a proof of concept and present it to the group when it's a little more developed.

xylar:
I had conversations with @mark-petersen and @matthewhoffman about this this morning. @matthewhoffman and I decided it would probably be okay to keep the tools here in `MPAS-Tools` and have a conda package that just pulls out a subset of those tools (e.g. mesh conversion tools and the new `planar_hex.py` python script, as well as some land-ice specific tools.

The main thing is that we would need to start making version tags (presumably releases) of MPAS-Tools so the conda package corresponds to a specific tag. @mgduda, any objections to this idea? It might be worth reconvening the MPAS-Dev team for a con call sometime soon to discuss this further.

PR #243, opened 2019-02-15
Create a conda recipe for the mesh conversion tools

xylar:
This will allow them to be installed as a conda package on any HPC machine (or at least most of them) without having to worry about installing the legacy NetCDF c++ library they depend on or building them from source.

This requires:
* bundling the legacy NetCDF C++ code with the mesh conversion tools
* modifying the Makefile to allow externally defined `CC` and `CFLAGS` environment variables
* removing libnetcdf-c++ from the build and instead pointing to the local cpp and header files.

A similar approach was taken by the [tempest-remap](https://github.com/ClimateGlobalChange/tempestremap) software.

PR #246, opened 2019-02-25
Add python code for computing planar hex meshes

xylar:
Currently produces bit-for-bit identical results to `periodic_hex/periodic_grid` but could be extended in the future to produce non-periodic meshes.

The reason for this update is that the python code will likely be much easier to incorporate into workflows like COMPASS because it doesn't require compilation or the Fortran NetCDF library.  It could also be incorporated quite easily into a conda package for mesh tools (see #243 and #245).

partially addresses #245

### DIFF
--- a/mesh_tools/planar_hex/planar_hex.py
+++ b/mesh_tools/planar_hex/planar_hex.py
@@ -1,0 +1,394 @@
+#!/usr/bin/env python
+
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
+import numpy
+import xarray
+import argparse
+import netCDF4
+
+
+def parse_args(mesh):
+    '''
+    Parse the command-line arguments and put them into the mesh as dimensions
+    or attributes.
+    '''
+
+    parser = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.RawTextHelpFormatter)
+    parser.add_argument('--nx', dest='nx', type=int, required=True,
+                        help='Cells in x direction')
+    parser.add_argument('--ny', dest='ny', type=int, required=True,
+                        help='Cells in y direction')
+    parser.add_argument('--dc', dest='dc', type=float, required=True,
+                        help='Distance between cell centers in meters')
+    parser.add_argument('-o', '--outFileName', dest='outFileName', type=str,
+                        required=False, default='grid.nc',
+                        help='The name of the output file')
+
+#    parser.add_argument('--periodicX', dest='periodicX', action='store_true',
+#                        help='Make the mesh periodic in x')
+#    parser.add_argument('--periodicY', dest='periodicY', action='store_true',
+#                        help='Make the mesh periodic in y')
+
+    args = parser.parse_args()
+
+    nx = args.nx
+    ny = args.ny
+    dc = args.dc
+
+    # non-periodic meshes aren't yet supported
+#    if args.periodicX:
+#        mesh.attrs['periodic_x'] = 'YES'
+#    else:
+#        mesh.attrs['periodic_x'] = 'NO'
+#    if args.periodicY:
+#        mesh.attrs['periodic_y'] = 'YES'
+#    else:
+#        mesh.attrs['periodic_y'] = 'NO'
+#    if args.periodicX or args.periodicY:
+#        mesh.attrs['is_periodic'] = 'YES'
+#    else:
+#        mesh.attrs['is_periodic'] = 'NO'
+
+    mesh.attrs['is_periodic'] = 'YES'
+    mesh.attrs['x_period'] = nx*dc
+    mesh.attrs['y_period'] = ny*dc*numpy.sqrt(3.)/2.
+
+    mesh.attrs['dc'] = dc
+
+    mesh.attrs['on_a_sphere'] = 'NO'
+    mesh.attrs['sphere_radius'] = 1.
+
+    nCells = nx * ny
+    nEdges = 3 * nCells
+    nVertices = 2 * nCells
+    vertexDegree = 3
+    maxEdges = 6
+
+    # add some basic arrays to get all the dimensions in place
+    indexToCellID = numpy.arange(nCells, dtype='i4')
+    indexToEdgeID = numpy.arange(nEdges, dtype='i4')
+    indexToVertexID = numpy.arange(nVertices, dtype='i4')
+
+    cellIdx = indexToCellID.reshape(ny, nx)
+    cellCol, cellRow = numpy.meshgrid(numpy.arange(nx, dtype='i4'),
+                                      numpy.arange(ny, dtype='i4'))
+
+    mesh['cellIdx'] = (('ny', 'nx'), cellIdx)
+    mesh['cellRow'] = (('nCells'), cellRow.ravel())
+    mesh['cellCol'] = (('nCells'), cellCol.ravel())
+
+    mesh['indexToCellID'] = (('nCells'), indexToCellID)
+    mesh['indexToEdgeID'] = (('nEdges'), indexToEdgeID)
+    mesh['indexToVertexID'] = (('nVertices'), indexToVertexID)
+
+    mesh['nEdgesOnCell'] = (('nCells',), 6*numpy.ones((nCells,), 'i4'))
+    mesh['cellsOnCell'] = (('nCells', 'maxEdges'),
+                           numpy.zeros((nCells, maxEdges), 'i4'))
+    mesh['edgesOnCell'] = (('nCells', 'maxEdges'),
+                           numpy.zeros((nCells, maxEdges), 'i4'))
+    mesh['verticesOnCell'] = (('nCells', 'maxEdges'),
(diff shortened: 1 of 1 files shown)


xylar:
I chose not to have the python code add the `Time` dimension or any tracers or vertical levels to the mesh file.  It looks like `periodic_hex` also stopped adding tracers and vertical levels but `namelist.input` still has options for them.  I don't believe the empty `Time` dimension is useful since no variables have that dimension.

I also chose not to produce the `graph.info.part.*` files.  My guess is that any workflow we have typically culls cells and then produces its own partition files.  But I would like to know if anyone uses that functionality and if I should add it back in.

matthewhoffman:
@xylar , I confirmed this works as advertised, and I agree with the decision to leave out Time and any non-mesh variables.  I also agree that there is no real value in the behavior of periodic_hex to output graph.info.part.* files, and the new script does not need to do that.  (The only really useful file is graph.info, but periodic_hex currently does not output that, and there is an existing script in the MPAS-Tools repo that can be used to create it.)  I also confirmed that this is substantially faster than the current code!

A couple of things I noticed:
1. Is requiring the xarray module necessary?  It seems like all functionality could be produced using just the netCDF4 module (which you already have as a requirement).

2. In periodic_hex, we added the requirement that ny be an even number, otherwise you get a mesh where the y-periodicity is messed up.  That should be added here too.

xylar:
@matthewhoffman, thanks for the review!

> 1. Is requiring the xarray module necessary?  It seems like all functionality could be produced using just the netCDF4 module

Yes, I rely extensively on xarray to handle the data structures and math between variables here.  So removing it would require a *substantial* rewrite, which I don't think is justified.  `xarray` is very easy to install with conda.  A lot of the speed-u likely comes because of `xarray`.

> 2. In periodic_hex, we added the requirement that ny be an even number, otherwise you get a mesh where the y-periodicity is messed up.  That should be added here too.

I must have missed that check.  I agree that it's necessary.

xylar:
I had a discussion with @matthewhoffman.  Both of us would like to know how others feel about the dependency on `xarray`.

milenaveneziani:
For me, it's a no brainer: if it works, and it's fast, and removing `xarray` means a substantial rewriting of the code, then why remove it??

matthewhoffman:
@xylar , I confirmed this works as advertised, and I agree with the decision to leave out Time and any non-mesh variables.  I also agree that there is no real value in the behavior of periodic_hex to output graph.info.part.* files, and the new script does not need to do that.  (The only really useful file is graph.info, but periodic_hex currently does not output that, and there is an existing script in the MPAS-Tools repo that can be used to create it.)  I also confirmed that this is substantially faster than the current code!

A couple of things I noticed:
1. Is requiring the xarray module necessary?  It seems like all functionality could be produced using just the netCDF4 module (which you already have as a requirement).

2. In periodic_hex, we added the requirement that ny be an even number, otherwise you get a mesh where the y-periodicity is messed up.  That should be added here too.

xylar:
Thanks or weighing in @milenaveneziani.  @mgduda, do you have an opinion about this?  Keep in mind that my primary reason for developing it is as a tool that can more easily be incorporated into a conda package (which can include xarray as a dependency without any extra pain for the user).  This need not be a replacement for `periodic_hex`, just an equivalent tool.

xylar:
@maltrud had raised the concern that xarray output wasn't working well in MPAS for him and @bradyrx.  I can confirm that I can run the results of this tool through `MpasMeshConverter.x` so it should at least be possible to start with this tool, use the mesh converter to "fix" it for MPAS if necessary, and proceed from there.  I haven't verified either way if I can run MPAS with the meshes this tool produces.  That'd definitely be a prerequisite of merging this PR.

bradyrx:
@xylar, to clarify, the issue I had was with using `xarray` to modify particle initialization files in LIGHT. For instance, using `xarray` to quickly filter out particles north of some latitude and then saving with `.to_netcdf()` would cause netCDF (I/O) errors when running MPAS. 

Even taking care to set the output format to NetCDF3_64BIT wouldn't help. I.e., metadata of the modified netCDF looked identical to that of files that worked with MPAS, but something under the hood from `xarray` caused bugs. Note that I did not try `MpasMeshConverter.x` to fix the bug. We went with the `netCDF4` package in python to fix the issues.

matthewhoffman:
@xylar , there was no assumption before that output from perdiodic_hex should work as a valid MPAS mesh without running through the mesh converter tool first.  So I think it's perfectly fine (and preferable) to still require that.

xylar:
@bradyrx, your case wouldn't have worked with the mesh converter, I suspect.  It's only for MPAS meshes, not *any* kind of MPAS data (e.g. particles).  I have had (minor) issues with fill values from xarray in my other work, which is why I have a helper function to use standard fill values in the output.  It just occurs to me that I have a run that's mysteriously crashing, and xarray could be the cause!  I wonder what it does that MPAS doesn't like.  Definitely worth looking into but it's outside the scope of this PR, for sure.

xylar:
> there was no assumption before that output from perdiodic_hex should work as a valid MPAS mesh without running through the mesh converter tool first. So I think it's perfectly fine (and preferable) to still require that.

@matthewhoffman, yep, that sound reasonable.  If results from `periodic_hex` didn't work without running mesh converter, there's no reason this tool would (or should) be any different.

matthewhoffman:
To clarify, the thinking Doug always had was to make only one tool (mesh converter) responsible for precise implementation of the MPAS mesh spec.  In the early days we kept running into issues where such and such tool or MPAS framework routine would appear broken only to discover some obscure detail of the mesh spec was not being obeyed in a mesh.  The mesh spec is complicated enough that requiring every tool to follow it to the letter isn't worth the effort.

xylar:
That's a good guiding principle, for sure.  It's obviously a little problematic that the tool for enforcing the mesh spec was written using c++, a language used by very few of the existing developers, and using a version of the NetCDF c++ library that was short-lived.  But hindsight is always 20/20 and we have ways of building that tool wherever needed now (#243).  For the record, I tried rewriting the mesh converter in python and it was a nightmare (and very slow) so I have a lot of respect of Doug's work on that tool.

mark-petersen:
@siddharthabishnu @amrapallig and @ipdemes This is a very useful tool that Xylar just added to create a 3D regular hex domain with your specifications.  Please give it a try.  It is in this branch of MPAS-Tools right now, but will be merged soon.  With this tool, you can do a spatial convergence test.

Here is a quick way to access it while it is on this branch:
```
git clone git@github.com:xylar/MPAS-Tools.git xylar-MPAS-Tools
cd xylar-MPAS-Tools
git checkout -b python_planar_hex origin/python_planar_hex
cd mesh_tools/planar_hex/
ls -l
-rwxr-xr-x 1 mpeterse mpeterse 16087 Feb 28 08:52 planar_hex.py
```
then use it with:
```
./planar_hex.py --nx 16 --ny 50 --dc 10000 -o base_mesh.nc
```
You will need the right python environment. On LANL IC:
```
module unload python; source /usr/projects/climate/SHARED_CLIMATE/anaconda_envs/load_latest_e3sm_unified.sh
```

If you have comments, or problems, you can post on this PR.

mark-petersen:
@siddharthabishnu @amrapallig and @ipdemes I should also say, after this command:
```
./planar_hex.py --nx 16 --ny 50 --dc 10000 -o base_mesh.nc
```
you still need to run the c-code mpas mesh converter:
```
./MpasMeshConverter.x base_mesh.nc mesh.nc
```
which produces the graph.info file. That code is housed here: https://github.com/MPAS-Dev/MPAS-Tools/tree/master/mesh_tools/mesh_conversion_tools.

@xylar Does planar_hex.py produce a valid MPAS mesh, or does `MpasMeshConverter` always need to be run as a second step?

xylar:
@mark-petersen, see discussion above.  @matthewhoffman thinks that no tool other than `MpasMeshConverter.x` should be trusted to produce a valid MPAS mesh, so we should always run any mesh through that tool.  I can't speak for whether things work without that for `planar_hex.py` but I certainly wouldn't guarantee it.

mgduda:
@matthewhoffman Actually, we do assume that the output of `periodic_hex` can be directly used by MPAS-Atmosphere for idealized simulations, with no need to run through a mesh converter tool. I can definitely help in testing that the output of the proposed Python tool works in this regard.

xylar:
@mgduda 
> I can definitely help in testing that the output of the proposed Python tool works in this regard.

That would be really great!  I'm a little worried about the NetCDF output from `xarray`, given @bradyrx's experience.

xylar:
@matthewhoffman, it sounds like you're in a minority in wanting the tool rewritten without `xarray`.  Given that and my reluctance to do so, are you willing to approve the PR as it is or would you like to see further changes?

matthewhoffman:
@xylar , if no one else objects to xarray, then I withdraw my concern.  Feel free to merge this!

xylar:
Thanks, @matthewhoffman.  I appreciate that your goal was to keep the tools in this repo as clean an simple to use as possible.  I'll definitely keep that goal in mind in my future additions.

PR #248, opened 2019-03-02
conda package for MPAS tools 

xylar:
This package is intended to include all the tools needed to run run COMPASS test cases and other typical mesh creation workflows.  This includes:
* the mesh conversion tools (#243)
* a python interface to the mesh conversion tools
* the planar hex creation tool (#246)
* a modified version of the mesh translation tool (python 3 compatible and callable as a function as well as a script
* 4 land-ice mesh tools (updated for python 3)
* 3 ocean coastal modification scripts (now available as functions and updated for python 3)
* add a copy of the MPAS-Model license (a license is required for conda-forge packages)
* a starting point for documentation

PR #249, opened 2019-03-02
Land ice mesh tools to python 3

xylar:
Make a few minor formatting changes to the land ice mesh tools so they work in python 3 as well as python 2.

I also include `mark_horns_for_culling.py` because I think this script might be used primarily by the land-ice model and because I'm pretty sure that @matthewhoffman is the author of all 5 scripts.

### DIFF
--- a/landice/mesh_tools_li/create_landice_grid_from_generic_MPAS_grid.py
+++ b/landice/mesh_tools_li/create_landice_grid_from_generic_MPAS_grid.py
@@ -3,6 +3,9 @@
 # I've only tested it with a periodic_hex grid, but it should work with any MPAS grid.
 # Currently variable attributes are not copied (and periodic_hex does not assign any, so this is ok).  If variable attributes are added to periodic_hex, this script should be modified to copy them (looping over dir(var), skipping over variable function names "assignValue", "getValue", "typecode").
 
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
+
 import sys, numpy
 from netCDF4 import Dataset
 from optparse import OptionParser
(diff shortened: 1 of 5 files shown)


xylar:
@matthewhoffman, here are changes needed to make the land ice mesh tools compatible with python 3.  This should give you a sense of some of the more basic changes that are typically required.  Mostly about print formatting.

A few of the formatting changes aren't  strictly necessary but there's a strong preference for the `.format()` as opposed to ` % ` or ` + ` and I was having to alter print statements all over in any case.  If you feel strongly about those changes, I can obviously revert them.

xylar:
I don't have a great way to test these.  I made sure they pass the `--help` test and that's about it.

xylar:
The intention is to revise these scripts before including them in the `mpas_mesh_tools` conda package (#248)

matthewhoffman:
@xylar , thanks for putting this together - it looks like a lot of tedious work.  I'll take a look and try the new versions out.

xylar:
@matthewhoffman, eh, not as much work as I was anticipating.  I'm starting to learn that a lot of the difficulties between python 2 and 3 come when you use fancier features like dictionaries and parse config files and stuff like that that you really only do when putting packages together, not scripts.  With scripts, it's almost entirely the print statements.

xylar:
Let me know if I should split `mark_horns_for_culling.py` into a separate PR.

xylar:
@matthewhoffman, could you suggest a process I could use to test these myself?  I would like to include them in #248, which I think @mark-petersen and I will work on getting merged this week.

mark-petersen:
My two cents: Let's take the leap on these and clean up any loose ends afterwards. @xylar has taken all reasonable precautions to change over to python 3.

matthewhoffman:
@xylar , that sounds entirely reasonable to me.  I apologize for my failing to give this the attention it deserves.

xylar:
Thanks @matthewhoffman.

PR #250, opened 2019-03-07
Coastline alteration tools to Python 3 and PEP8

xylar:
This merge updates to python 3 with PEP8 formatting the two older python scripts used for coastal alternation for ocean meshes (via COMPASS).


xylar:
@mark-petersen, this is a small part of my ongoing work toward a conda package that will contain all the tools COMPASS needs.  In that process, I'm making sure everything should work with python 3, as here.

PR #252, opened 2019-03-26
Add Tool to merge 2 MPAS non-contiguous meshes into a single file

matthewhoffman:
This tool takes two MPAS meshes and combines them into a single file.  The two meshes have no topological connections in the new file, but both regions co-exist.  The two regions may overlap in space.

### DIFF
--- a/mesh_tools/merge_split_meshes/merge_grids.py
+++ b/mesh_tools/merge_split_meshes/merge_grids.py
@@ -1,0 +1,191 @@
+#!/usr/bin/env python
+'''
+Tool to merge 2 MPAS non-contiguous meshes together into a single file
+'''
+
+import sys
+import numpy as np
+import netCDF4
+import argparse
+import math
+from collections import OrderedDict
+import scipy.spatial
+import time
+from datetime import datetime
+
+
+#print "== Gathering information.  (Invoke with --help for more details. All arguments are optional)\n"
+parser = argparse.ArgumentParser(formatter_class=argparse.RawTextHelpFormatter)
+parser.description = __doc__
+parser.add_argument("-1", dest="file1", help="name of file 1", metavar="FILENAME")
+parser.add_argument("-2", dest="file2", help="name of file 2", metavar="FILENAME")
+parser.add_argument("-o", dest="outFile", help="name of output file", default="merged_mpas.nc", metavar="FILENAME")
+#for option in parser.option_list:
+#    if option.default != ("NO", "DEFAULT"):
+#        option.help += (" " if option.help else "") + "[default: %default]"
+options = parser.parse_args()
+
+
+
+f1 = netCDF4.Dataset(options.file1)
+nCells1 = len(f1.dimensions['nCells'])
+nEdges1 = len(f1.dimensions['nEdges'])
+nVertices1 = len(f1.dimensions['nVertices'])
+Time1= len(f1.dimensions['Time'])
+
+f2 = netCDF4.Dataset(options.file2)
+nCells2 = len(f2.dimensions['nCells'])
+nEdges2 = len(f2.dimensions['nEdges'])
+nVertices2 = len(f2.dimensions['nVertices'])
+Time2= len(f2.dimensions['Time'])
+
+if Time1 != Time2:
+  sys.exit("ERROR: The two files have different lengths of the Time dimension.")
+if len(f1.dimensions['vertexDegree']) != len(f2.dimensions['vertexDegree']):
+  sys.exit("ERROR: The two files have different lengths of the vertexDegree dimension.")
+if len(f1.dimensions['nVertLevels']) != len(f2.dimensions['nVertLevels']):
+  sys.exit("ERROR: The two files have different lengths of the nVertLevels dimension.")
+
+
+# Create new file
+fout = netCDF4.Dataset(options.outFile, "w", format="NETCDF3_CLASSIC")
+
+# add merged dimensions
+print("Adding merged dimensions to new file.")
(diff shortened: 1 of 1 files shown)


matthewhoffman:
@mgduda , @xylar , @pwolfram , @akturner, @mark-petersen , others, let me know if any of you are interested in reviewing this.  It is kind of an obscure use-case, but it may be useful to other cores.

This image shows an example of merging planar Greenland and Antarctica meshes into one netCDF file:
Screen Shot 2019-03-26 at 10 55 33 AM

mesh_tools/merge_split_meshes/merge_grids.py
```diff
@@ -0,0 +1,191 @@
...
+nEdges2 = len(f2.dimensions['nEdges'])
+nVertices2 = len(f2.dimensions['nVertices'])
+Time2= len(f2.dimensions['Time'])
+
+if Time1 != Time2:
+  sys.exit("ERROR: The two files have different lengths of the Time dimension.")
+if len(f1.dimensions['vertexDegree']) != len(f2.dimensions['vertexDegree']):
+  sys.exit("ERROR: The two files have different lengths of the vertexDegree dimension.")
+if len(f1.dimensions['nVertLevels']) != len(f2.dimensions['nVertLevels']):
+  sys.exit("ERROR: The two files have different lengths of the nVertLevels dimension.")
```
xylar: Would it make sense to first find out if `nVertLevels` is even in the meshes? I was playing with a mesh that doesn't have that dimension and I think there are other components that don't support that dimension.

mesh_tools/merge_split_meshes/merge_grids.py
```diff
@@ -0,0 +1,191 @@
...
+nCells2 = len(f2.dimensions['nCells'])
+nEdges2 = len(f2.dimensions['nEdges'])
+nVertices2 = len(f2.dimensions['nVertices'])
+Time2= len(f2.dimensions['Time'])
+
+if Time1 != Time2:
+  sys.exit("ERROR: The two files have different lengths of the Time dimension.")
+if len(f1.dimensions['vertexDegree']) != len(f2.dimensions['vertexDegree']):
+  sys.exit("ERROR: The two files have different lengths of the vertexDegree dimension.")
+if len(f1.dimensions['nVertLevels']) != len(f2.dimensions['nVertLevels']):
```
xylar:
```suggestion
if 'nVertLevels' in f1.dimensions and 'nVertLevels' in f2. dimensions and \
          len(f1.dimensions['nVertLevels']) != len(f2.dimensions['nVertLevels']):
```

mesh_tools/merge_split_meshes/merge_grids.py
```diff
@@ -0,0 +1,191 @@
...
+fout.createDimension('nEdges', nEdges1+nEdges2)
+fout.createDimension('nVertices', nVertices1+nVertices2)
+fout.createDimension('TWO', 2)
+fout.createDimension('vertexDegree', len(f1.dimensions['vertexDegree']))
+if 'StrLen' in f1.dimensions:
+   fout.createDimension('StrLen', len(f1.dimensions['StrLen']))
+maxEdges = max(len(f1.dimensions['maxEdges']), len(f2.dimensions['maxEdges']))
+fout.createDimension('maxEdges', maxEdges)
+fout.createDimension('maxEdges2', maxEdges*2)
+fout.createDimension('nVertLevels', len(f1.dimensions['nVertLevels']))
```
xylar:
```suggestion
if 'nVertLevels' in f1.dimensions:
    fout.createDimension('nVertLevels', len(f1.dimensions['nVertLevels']))
```

mesh_tools/merge_split_meshes/merge_grids.py
```diff
@@ -0,0 +1,191 @@
...
+fout.createDimension('nVertInterfaces', len(f1.dimensions['nVertInterfaces']))
```
xylar: if 'nVertInterfaces' in f1.dimensions:
    fout.createDimension('nVertInterfaces', len(f1.dimensions['nVertInterfaces']))

xylar:
I tested in python 3.7, so I'm adding a "python3" label as a stamp of approval.

matthewhoffman:
@xylar , thanks for looking it over and testing in python 3.  Does this not require any of the 'from futures' business, because none of that changing functionality (e.g., division) is being used?

I will do the clean-up you suggest (as well as copying, rather than hard-coding, the ON_A_SPHERE and related attributes).

xylar:
> Does this not require any of the 'from futures' business, because none of that changing functionality (e.g., division) is being used?

Yes, that's basically right.  If you add those things, it will make python 2 act a bit more like python 3.  For example, it would prevent you from using print statements without parentheses.  But for relatively simple scripts like this one it's not necessary.

xylar:
Can you imagine a case where this would work with periodic meshes?  I can't imagine merging 2 periodic meshes and having them still be valid periodic meshes.  So I wouldn't copy the `is_periodic` attribute and leave it hard-coded to `"NO"`.

xylar:
It might even be worth barfing if either input mesh is periodic, except that we usually forget to fix the attribute when we cull cells to make meshes not be periodic.

mgduda:
@matthewhoffman Recognizing that the script does have a `--help` option, would it nonetheless be worth adding an error message if the script is run without the required number of command-line arguments?

mesh_tools/merge_split_meshes/merge_grids.py
matthewhoffman: Good suggestion - added to new commit.

PR #255, opened 2019-05-14
Fix vtk extractor bug

bradyrx:
Quick fix on a bug @pwolfram and I found in `paraview_vtk_field_extractor.py`. 

Note that `time_series_file` is closed, and then called in the else block in L175. This breaks the extractor if one just uses the `-f` flag (and not the `-m` flag), e.g. when the mesh and desired variables are contained in the same file.

https://github.com/MPAS-Dev/MPAS-Tools/blob/e2202ecb0a98562a9221971f8dd01dd4c5fc08c8/visualization/paraview_vtk_field_extractor/paraview_vtk_field_extractor.py#L167-L177

Although note that `time_series_file.close()` is called again towards the end of this block:
https://github.com/MPAS-Dev/MPAS-Tools/blob/e2202ecb0a98562a9221971f8dd01dd4c5fc08c8/visualization/paraview_vtk_field_extractor/paraview_vtk_field_extractor.py#L373

So perhaps this is a merge bug and it should just remain at L373 and be removed around the region I implemented the fix.

### DIFF
--- a/visualization/paraview_vtk_field_extractor/paraview_vtk_field_extractor.py
+++ b/visualization/paraview_vtk_field_extractor/paraview_vtk_field_extractor.py
@@ -164,8 +164,6 @@ def build_field_time_series(local_time_indices, file_names, mesh_file,
         else:
             nHyperSlabs += len(extra_dim_vals)
 
-    time_series_file.close()
-
     any_var_has_time_dim = np.any(var_has_time_dim)
 
     if topo_dim is not None:
@@ -175,6 +173,8 @@ def build_field_time_series(local_time_indices, file_names, mesh_file,
             nTopoLevels = len(time_series_file.dimensions[topo_dim])
     else:
         nTopoLevels = None
+
+    time_series_file.close()
 
     try:
         os.makedirs(out_dir)


xylar:
@bradyrx, this looks good to me.  Can you explain why there's a weird merge commit in this PR, though?  Seems like that could be removed with a rebase.

bradyrx:
@xylar, the merge commit is because I had an old MPAS-Tools fork and caught it up to master from the main repo. It appended onto this PR for some reason.

I'm having trouble getting rid of it through my standard method. Doing the following:
```shell
git clone git@github.com:bradyrx/MPAS-Tools.git
cd MPAS-Tools
git remote add main git@github.com:MPAS-Dev/MPAS-Tools.git
git fetch --all
git checkout fix_vtk_extractor_bug
git rebase -i main/master
```

I am only left with the following commit, without the merge commit to fixup.
`>>> pick c9e34e3 bugfix on time_series_file.close() location`

Following rebase the log shows the merge commit disappeared. So I run `git push -f origin/fix_vtk_extractor_bug` per usual to update the commits, and get `fatal: You didn't specify any refspecs to push, and push.default is "nothing".`

Any thoughts on how to drop that merge fix? I could open a new PR without it, but this seems like a good learning opportunity.

xylar:
@bradyrx, the only problem is your push:
```
git push -f origin fix_vtk_extractor_bug
```
You had a slash there that shouldn't be there.

bradyrx:
Sigh. I was 95% of the way there. Thanks @xylar. This should be good to go!

xylar:
Thanks, @bradyrx!

PR #256, opened 2019-05-17
Add an external script to conda-package planar_hex

xylar:
In #248, `planar_hex` was moved into the conda-package for MPAS-Tools.  However there is also a need for an external version of the tool.  This PR adds a stub script called planar_hex that calls the code in the conda package.  (The stub script is copied from a conda environment and is the same approach used to access entry points into python packages used elsewhere.)

### DIFF
--- a/mesh_tools/planar_hex/mpas_tools
+++ b/mesh_tools/planar_hex/mpas_tools
@@ -1,0 +1,1 @@
+../../conda_package/mpas_tools/
--- a/mesh_tools/planar_hex/planar_hex
+++ b/mesh_tools/planar_hex/planar_hex
@@ -1,0 +1,11 @@
+#!/usr/bin/env python
+
+# -*- coding: utf-8 -*-
+import re
+import sys
+
+from mpas_tools.planar_hex import main
+
+if __name__ == '__main__':
+    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
+    sys.exit(main())


xylar:
@matthewhoffman, could you see if this works for you?

xylar:
@matthewhoffman, my preferred way of having `planar_hex` used would be for you to either use the `compass_py3.7` environment that I set up on IC:
```
source /usr/projects/climate/SHARED_CLIMATE/anaconda_envs/base/etc/profile.d/conda.sh
conda activate compass_py3.7
```
or if there are packages you need that aren't in that environment, include `mpas_tools` in your own conda environment:
```
conda create -n my_awesome_env -c conda-forge python=3.7 mpas_tools numpy scipy netcdf4 ...
conda activate my_awesome_env
```

But I think it makes sense to have the option to just run `planar_hex` it out of the repo for those with a different (dare I say outdated) workflow.

matthewhoffman:
@xylar , thanks for addressing this.  Frankly, I'd be ok with moving to the conda env approach to using tools in a workflow - I just need to get myself up to speed on that.    I'm curious about why this tool ended up in the conda directory whereas all the others are still standalone scripts, but I can probably wait until next time I see you to get the explanation.

xylar:
>  I'm curious about why this tool ended up in the conda directory whereas all the others are still standalone scripts, but I can probably wait until next time I see you to get the explanation.

I'd personally move all the tools to the package but I don't feel like I "own" any of them but the ones I wrote recently.  I figured `planar_hex` as new enough that no one had really had time to work it into their workflows, whereas others would take awhile before the conda environment was the primary way that they're used.  I essentially gutted a lot of the ocean tools and just left stub scripts behind, a bit like what I've now done for `planar_hex` but that approach is clumsy and it's better if people know to look in the conda package instead.  Ideally, I'd like to do the same with land-ice tools but the only think that's really required is converting them to python 3.

xylar:
@matthewhoffman, could you test and approve this at your convenience so I can merge it?

xylar:
Great, thanks for trying it out and for the helpful discussion.

PR #257, opened 2019-05-29
Merge Split tool; an addition to PR #252

jhkennedy:
PR #252 introduces a mesh merge tool. Here, we add a `split_grids.py` tool which can be used to reverse the process. I've tested it with the `aisgis20km.20190326.nc` (create via `merge_grids.py`) and am able to reproduce exactly the `ais20km.150910.nc` and `gis20km.150922.nc` meshes.


jhkennedy:
@matthewhoffman and @xylar I can't suggest reviewers for this PR, but you'll both definitely want to look it over. 

@matthewhoffman Of particular interest to #252 , if you add a `merge_point` global attribute like so: 

```py
fout.merge_point = json.dumps({'nCells': nCells1, 
                               'nEdges': nEdges1, 
                               'nVertices': nVertices1})
```
this tool can automatically split the grids back out. 

Also, because of [this bit](https://github.com/MPAS-Dev/MPAS-Tools/pull/252/files#diff-a9ab7154d1f6de62c1a2355b74f217e6R62) in `merge_grids.py` :

```py
maxEdges = max(len(f1.dimensions['maxEdges']), len(f2.dimensions['maxEdges']))
```
I won't be able to get exactly the same `maxEdges` as before if they differed. The `merge_point` spec above could be amended to include that info and then I should be able to completely specify it.

jhkennedy:
Also, the is a little bit of boilerplate in `split_grids.py` so that it could be used progromatically like:

```py
# Assuming it's on the python path...
import split_grids
arg_list = ['in.nc', '-1', 'out1.nc', '-2', 'out2.nc']
split_grids.main(arg_list)
```

jhkennedy:
Also, also, I can put the `merge_point` attribute in `merge_grid.py` and similar boilerplate so it can be used programmatically, but I don't want to make changes to `merge_grid.py` until #252 is merged in so this PR diff shows those changes.

matthewhoffman:
@jhkennedy , this looks great.  I'll prioritize finishing #252 so that your PR can get finished!

matthewhoffman:
@jhkennedy , I merged #252 after some minor revisions, including adding the merge_point data, so you can go ahead and rebase this PR and make the additional changes you wanted to make.  They sound good to me.  Feel free to edit the structure of the merge_point attribute as needed/desired.

xylar:
@jhkennedy, this is  nice feature.  I would personally handle the case where you want to import this and call it as a function a little differently.  Namely, I would want to be able to do something like:
```python
# Assuming it's on the python path...
from split_grids import split_grids
split_grids(infile='in.nc', outfile1='out1.nc', outfile2='out2.nc')
```
with normal python kwargs type syntax.  Then, you could assume your `main()` only gets called when the function is run as a script and parse arguments there before calling `split_grids()`.  Obviously, this is a question of taste but I don't think the list of command-line arguments getting passed to main is very convenient or intuitive when importing and calling the `main()` function directly.

xylar:
Also, should these merge and split tools be added to the conda package? If so, maybe another PR is the right place to add them.

jhkennedy:
@xylar you're right, that's a much better API, I'll make that change. And yeah, at least the split_mesh's tool I'd like in the conda package as I expect one or all of Processflow, LIVVkit, and MPAS-Analysis would take advantage of it when looking at ice sheet SMB (at least for the time being until multiple ice sheets are able to run in E3SM).

As for doing conda here, or in another PR, I think that's up to you and the other MPAS-Tools devs; it doesn't matter to me either way.

xylar:
Okay, in that case I'd suggest you add both the merging and splitting tools to the conda package here.  How you do that it up to you.  With some tools I've been integrating them as entry points into:
https://github.com/MPAS-Dev/MPAS-Tools/tree/master/conda_package/mpas_tools
For others, I've been leaving them in place for legacy reasons, but just as a stub that calls into functions that are part of the python package, see these examples:
https://github.com/MPAS-Dev/MPAS-Tools/tree/master/ocean/coastline_alteration

I then copy these scripts into subdirectories of `conda_package` during the conda-forge build:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/conda_package/recipe/build.sh#L6
and add them as scripts to `setup.py`:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/conda_package/setup.py#L31-L40
which is obviously kind of a cluge.

I'm more than happy to hear your feedback on other aspects of the conda package (which is still messy and kind of a work in progress), but maybe better to keep that as a separate issue or PR.

mesh_tools/merge_split_meshes/split_grids.py
```diff
@@ -0,0 +1,235 @@
...
+            slice2 += (slice(0, maxEdges[1]*2),)
+        else:
+            slice1 += (slice(None),)
+            slice2 += (slice(None),)
+
+    return slice1, slice2
+
+
+# NOTE: Python 2 and 3 string fun conflicting with NC_CHAR vs NC_STRING, see:
+#       https://github.com/Unidata/netcdf4-python/issues/529
```
xylar: Pain in the a$$!

matthewhoffman:
In py2.7 I got this error:
```
Traceback (most recent call last):
  File "./merge_grids.py", line 217, in <module>
    arguments = parse_args()
  File "./merge_grids.py", line 29, in parse_args
    return parser.parse_intermixed_args(args)
AttributeError: 'ArgumentParser' object has no attribute 'parse_intermixed_args'
```
but it sounds like we should no longer worry about supporting py2.7, so that's ok.  @xylar , do you agree?

jhkennedy:
@matthewhoffman ah yeah, I've not tested yet with python 2. @xylar should we support 2? If so, I can switch it to `parse_args` -- the only difference being the options must come before the required args in that case.

jhkennedy:
@matthewhoffman Thanks! I'm making some changes to move these scripts into the conda package as well

xylar:
@jhkennedy and @matthewhoffman, I don't think this PR is the right one to use to drop python 2.7 support.  I agree that's the direction we're moving but so far we have no python 3 only scripts and many python 2 only scripts.  I would say everything needs to support both at least a little while longer.

xylar:
> many python 2 only scripts

These are only *outside* of the package, but the package gets build on conda-forge as noarch so people might assume it supports python 2.7.

conda_package/docs/api.rst
```diff
@@ -14,7 +14,7 @@ MPAS mesh tools
 .. autosummary::
    :toctree: generated/
 
-   make_periodic_planar_hex_mesh
+   make_planar_hex_mesh
```
xylar: Oops, thanks!

conda_package/mpas_tools/__init__.py
```diff
@@ -1,2 +1,2 @@
-__version_info__ = (0, 0, 1)
+__version_info__ = (0, 0, 2)
```
xylar: Again, oops, thanks!

jhkennedy:
@matthewhoffman and @xylar, I've moved `merge_grids` and `split_grids` into the conda package and left a symbolic link to them in `/mesh_tools/merge_split_meshes/` so that they can be used from there without having to install the conda package. Also, I've now tested with both python 2.7 and 3.7, so it should be compatible. 

Unless you see anything else, I think it's ready!

xylar:
Yep, still looks good to me.  @matthewhoffman, merge when you're ready.

xylar:
@matthewhoffman, is this ready to merge in your view?  If so, I would like to include it in v0.0.3 of the conda package.

xylar:
No need to worry about testing the conda package, just the "standalone" functionality.  I will test the conda package in #261 and make changes there if necessary.

matthewhoffman:
@xylar , yes this is good - I've merged it so it will be included in the conda update.

PR #259, opened 2019-06-13
Fix writing of pvd files in the paraview extractor

### LINKED ISSUE #258: paraveiw extractor producing empty pvd files
I haven't had time to investigate this further but it seems like either recent changes or the use of python 3 has led to the extractor producing empty pvd files.  This needs to be fixed!

xylar:
The files were not being explicitly closed, which may only be an issue in python 3.

This merge also fixes some formatting issues that violated PEP8.

closes #258


xylar:
@mark-petersen, could you try this out quickly when you get a chance?  I just need you to run the extractor as you normally would (but make sure it's from a python 3 environment, since I think that's where the issue shows up).  If you get a non-empty pvd file, we're in business!

PR #260, opened 2019-06-14
Convert remaining landice-related tools to py3

matthewhoffman:
This merge converts the remaining landice tools to py3, as well as some additional shared tools that landice COMPASS uses.  It also adds a couple shared tools that landice COMPASS requires to the conda package.

PR #261, opened 2019-06-15
Update conda package to v0.0.3

xylar:
...for a release and build on conda-forge.


xylar:
To be merged after #260, just before making a new release.

xylar:
@matthewhoffman and @pwolfram, I'm hoping to make this release relatively soon so I can include it in the next deployment of e3sm-unified.  @matthewhoffman, do you have an estimated time table for when you'll be ready to merge #260?  @pwolfram, were you able to figure out if there are scripts that you would like to have added to the conda package, and if so do you have a time table for when you can do a PR to add them?

xylar:
@jhkennedy, I will ask you to give this a quick look once I rebase after #260 gets merged.

jhkennedy:
@xylar Happy to!

pwolfram:
@xylar, thanks for the offer-- I won't have time this time to do this in the short term.  Long term, however, I would argue we want the JIGSAW scripts to be in the conda build, which doesn't appear to be the case.  The key use case in my mind is simplification of COMPASS or tools we need to use frequently and want to have on the command line.  Are there other use cases?  

Thanks for the offer!  Maybe we should submit an issue for discussion on migration of the JIGSAW mesh conversion scripts?

cc @dengwirda, @kristin-hoch, @sbrus89

matthewhoffman:
@xylar , I'll try to get #260 tested and merged today.  I may need to try building the package on a different computer.

Thanks for putting together a conda package update.  That will help with the COMPASS changes I need to do.

xylar:
> Long term, however, I would argue we want the JIGSAW scripts to be in the conda build, which doesn't appear to be the case.

To be clear, mpas_tools is not meant to supply all the packages needed by compass.  I have a separate readme that defines those needs:

https://github.com/MPAS-Dev/MPAS-Model/blob/ocean/develop/testing_and_setup/compass/README_ocean.md

So far, there is no compelling reason to make a conda package for compass, since we always need to run out of the repo.

> The key use case in my mind is simplification of COMPASS or tools we need to use frequently and want to have on the command line. Are there other use cases?

That's the purpose of the compass conda environment, which is a super-set of mpas_tools and several other packages mentioned in the README.

> Thanks for the offer! Maybe we should submit an issue for discussion on migration of the JIGSAW mesh conversion scripts?

I think those belong in COMPASS, since they don't have any purpose outside of making test cases.  They could be moved here but I don't see a big advantage.

xylar:
> I'll try to get #260 tested and merged today. I may need to try building the package on a different computer.

No rush, just wanted a sense of it.

> Thanks for putting together a conda package update. That will help with the COMPASS changes I need to do.

Oh, sure, no problem!

xylar:
@jhkennedy, I'm doing a test build now.  I'd appreciate a quick review.  Mostly just a sanity check to make sure I update the version number, dependencies, tests, etc. as needed.

conda_package/recipe/meta.yaml
```diff
@@ -26,6 +26,7 @@ requirements:
     - hdf5
     - libnetcdf
     - setuptools
+    - netcdf4
```
xylar: This may not be necessary anymore but I'm leaving it for now.

conda_package/recipe/meta.yaml
```diff
@@ -26,6 +26,9 @@ requirements:
...
+    - openmp  # [osx]
```
xylar: This was necessary on conda-forge.  @matthewhoffman, this might fix the compiler error you saw.  Fingers crossed...

xylar:
Thanks @jhkennedy.  I've made all your suggested changes.  We don't need the reg. ex. version in `conf.py` because we import `mpas_tools` in any case to do the auto-generated documentation.  But for `setup.py`, that's super helpful.

PR #269, opened 2019-10-03
Fix conda package setup.py when called with a path

xylar:
In such cases, we need to cd to the location of `setup.py` before finding package modules and scripts.

This appears to be needed to build in Read The Docs.


xylar:
Sorry everyone for these frequent merges but Read The Docs is hard to debug any other way.

PR #270, opened 2019-10-03
In conda package, copy directories within setup.py

xylar:
This was being done externally in scripts but it is easier to include it in `setup.py`.  This also seems to be the only way to do this within Read The Docs, which doesn't support custom install scripts.

### DIFF
--- a/conda_package/setup.py
+++ b/conda_package/setup.py
@@ -3,6 +3,7 @@
 import os
 import re
 from setuptools import setup, find_packages
+import shutil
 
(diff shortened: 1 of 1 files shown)


xylar:
Hopefully this is the last of these annoying PRs...

xylar:
It build, it build! https://mpas-tools.readthedocs.io/en/latest/

I'll post this in the `README.md` once we've built it out enough to actually be of some use.

PR #271, opened 2019-10-28
Upgrade create_SCRIP_file_from_MPAS_mesh.py to python3

mark-petersen:
Simple upgrade to python 3. Just print parenthesis and spacing.

### DIFF
--- a/mesh_tools/create_SCRIP_files/create_SCRIP_file_from_MPAS_mesh.py
+++ b/mesh_tools/create_SCRIP_files/create_SCRIP_file_from_MPAS_mesh.py
@@ -9,7 +9,7 @@ import numpy as np
 from optparse import OptionParser
 
 
-print ("== Gathering information.  (Invoke with --help for more details. All arguments are optional)")
+print("== Gathering information.  (Invoke with --help for more details. All arguments are optional)")
 parser = OptionParser()
 parser.description = "This script takes an MPAS grid file and generates a SCRIP grid file."
 parser.add_option("-m", "--mpas", dest="mpasFile", help="MPAS grid file name used as input.", default="grid.nc", metavar="FILENAME")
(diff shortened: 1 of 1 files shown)


mark-petersen:
Tested with e3sm-unified package on grizzly. It works.

xylar:
@mark-petersen, could you run autopep8 on the script.  This will take out the spaces in your various `print ()` statements and may find some other issues.

Also, would it make sense to add this script to the conda package?  My guess is that it would, since we might want to use it in COMPASS.

xylar:
I take that back, don't run autopep8.  I tried it and it changes too much.
Instead, could you just search and replace `print (` with `print(`?

xylar:
@matthewhoffman, do you want to make `create_SCRIP_file_from_CISM_mesh.py` work with python 3 as part of this PR?

xylar:
It looks like the conda package needs some attention so let me take care of adding the script to the package in a separate PR.

mark-petersen:
@xylar thanks for looking at this. Funny, the space in `print ("` reveals that I think of print as a statement rather than a function. Old habits die hard. Lillian teases me that I use two spaces after a period when typing, which reveals that I am an "old person".

xylar:
> Funny, the space in print (" reveals that I think of print as a statement rather than a function. 

That's precisely what changed from python 2.7 to 3: `print` became a function and that's why you now need the parentheses.

> Lillian teases me that I use two spaces after a period when typing, which reveals that I am an "old person".

I do that, too.  I learned to type properly on a typewriter in high school.  I'll never be able to break the habit now.  Hopefully the current generation of postdocs are young enough to not have picked up that habit.

xylar:
@mark-petersen, your call if you want to wait on @matthewhoffman to see if he wants to fix the CISM script or if we just let this be.

mark-petersen:
Merging. @matthewhoffman feel free to alter the land ice script when you see fit.

xylar:
@mark-petersen, please include the PR number in your merge commit.  It helps a lot with drafting a release.  The easiest way to do that is to just use the merge button here on GitHub.

mark-petersen:
OK, good to know. I never use the button on the main repo because I always merge locally and test again. Do you want me to change the text on the merge commit and do a force push?

xylar:
@mark-petersen, no, I don't think we should *ever* do a force push to master.  Plus, I've already done the release.

I really think the merge button is quite safe.  I understand that it's the policy not to use it on the MPAS-Model repo (though I think that's unnecessarily cautious) but I have never had problems with it.  It always does a merge commit rather than a fast-forward, as we would want.  My understanding was that was always the concern with MPAS-Model.

PR #272, opened 2019-10-28
Updates to the conda package for release of v0.0.4

xylar:
* Add SCRIP script to package
* fix some tests
* fix copying of scripts/code from outside of the package
* update to v0.0.4


xylar:
I did a local test build of v0.0.4 and everything went okay, so I think this can be released.

PR #277, opened 2019-12-20
Add planar_hex args as global attributes mesh.nc

mark-petersen:
It is useful to include the flag settings nx,ny,dc in the flag settings. When I plot regular planar meshes, it is convenient to grab nx by ny back out for a reshape. If people don't use these global attributes, there is no harm done.


mark-petersen:
I tested this, and it works with no problem. There is no rush, can be included in the next regular conda package release.

PR #279, opened 2020-01-22
Fixes output of x_period and y_period

vanroekel:
For periodic grids, x and y_period where being read in as character
strings and then converted to doubles for output in the mesh.  The read
in as a string caused only the first two decimals to be read of y_period
and x_period, which breaks periodicity in MPAS.

This commit reads in x_period and y_period as doubles and skips the
conversions.

### DIFF
--- a/mesh_tools/mesh_conversion_tools/netcdf_utils.cpp
+++ b/mesh_tools/mesh_conversion_tools/netcdf_utils.cpp
@@ -449,9 +449,8 @@ double netcdf_mpas_read_yperiod(string filename){/*{{{*/
 		vals = att_id -> values();
 
 		for(int i = 0; i < vals -> num(); i++){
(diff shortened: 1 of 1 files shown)


vanroekel:
I've confirmed this fix restores doubly periodic behavior for my COMPASS test case. I can show before and after if you'd like

xylar:
@vanroekel, do you want a new release of the mpas-tools package as soon as this goes in?  That would allow us to update the COMPASS environments but would still take awhile to make its way to E3SM-Unified.

vanroekel:
@xylar having a new release of the compass environment would be much appreciated.

xylar:
@pwolfram, can you review this as soon as you're able so I can get this in?

pwolfram:
@xylar, thanks for your patience-- it's all good to go!

PR #282, opened 2020-01-27
Fixes landice mesh tool interpolate_to_mpasli_grid.py when source and destination grids are both mpas format.

trhille:
Used input file for vertical dimensions, when should have used output file.

### DIFF
--- a/landice/mesh_tools_li/interpolate_to_mpasli_grid.py
+++ b/landice/mesh_tools_li/interpolate_to_mpasli_grid.py
@@ -366,7 +366,7 @@ def interpolate_field_with_layers(MPASfieldName):
 def vertical_interp_MPAS_grid(mpas_grid_input_layers, input_layers):
     destinationField = np.zeros((nCells, nVertLevels))
     for i in range(nCells):
-        destinationField[i,:] = np.interp(mpasLayerCenters, input_layers, mpas_grid_input_layers[:,i])
+        destinationField[i,:] = np.interp(mpasLayerCenters,input_layers, mpas_grid_input_layers[:,i])
     return destinationField
 
 
@@ -478,7 +478,7 @@ if filetype=='cism':
 elif filetype == 'mpas':
     # Get the MPAS vertical dimensions if they exist
     try:
-      nVertLevels = len(inputFile.dimensions['nVertLevels'])
+      input_nVertLevels = len(inputFile.dimensions['nVertLevels'])
     except:
       print('  Input file is missing the dimension nVertLevels.  Might not be a problem.')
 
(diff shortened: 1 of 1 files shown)


xylar:
@trhille, it would be helpful for other users of MPAS-Tools if the name of the PR had some information about which component and/or tool is being modified.  But that way folks know whether it's relevant to them or not.  Easy enough to figure out once we go to the PR, but we're lazy ;-)

matthewhoffman:
@xylar , you can blame for poor guidance on that.  That's a good point.  (@trhille , you can go in and edit the existing title from this page.)

trhille:
@xylar @matthewhoffman Thanks. Should be better now.

xylar:
Nice!  And no blame, just figured it would be helpful to know for future PRs, too.

landice/mesh_tools_li/interpolate_to_mpasli_grid.py
```diff
@@ -478,7 +478,7 @@ def vertical_interp_MPAS_grid(mpas_grid_input_layers, input_layers):
 elif filetype == 'mpas':
     # Get the MPAS vertical dimensions if they exist
     try:
-      nVertLevels = len(inputFile.dimensions['nVertLevels'])
+      input_nVertLevels = len(inputFile.dimensions['nVertLevels'])
```
matthewhoffman: @trhille , looking at this closer, this dimension variable is not actually used anywhere in the script, so a cleaner solution would be to simply delete this entire try/except clause (i.e. I never should have added it in the first place).  (The interpolation part figures out the vertical information for the source file around lines 280-291.)  You can delete these lines in a new commit and just push it to your branch, and the PR will update automatically.

trhille:
@matthewhoffman Thanks. Done.

landice/mesh_tools_li/interpolate_to_mpasli_grid.py
```diff
@@ -366,7 +366,7 @@ def interpolate_field_with_layers(MPASfieldName):
 def vertical_interp_MPAS_grid(mpas_grid_input_layers, input_layers):
     destinationField = np.zeros((nCells, nVertLevels))
     for i in range(nCells):
-        destinationField[i,:] = np.interp(mpasLayerCenters, input_layers, mpas_grid_input_layers[:,i])
+        destinationField[i,:] = np.interp(mpasLayerCenters,input_layers, mpas_grid_input_layers[:,i])
```
xylar: I'm guessing this change was inadvertent?

matthewhoffman: Yeah, and I was ignoring it in the attempt to keep things simple for an intro to PRing.

PR #284, opened 2020-02-05
Add 3D capability to landice mesh tool conversion_exodus_init_to_mpasli_mesh.py

trhille:
Allows for conversion of temperature, velocity, and other 3D fields from exodus to the mpas mesh.

### DIFF
--- a/landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
+++ b/landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
@@ -65,8 +65,8 @@ ice_density = 910.0
 ocean_density = 1028.0
 
 dataset = Dataset(options.nc_file, 'r+')
-x = dataset.variables['xCell'][:]
-y = dataset.variables['yCell'][:]
+xCell = dataset.variables['xCell'][:]
+yCell = dataset.variables['yCell'][:]
 
 exo = exodus(options.exo_file)
 
@@ -214,17 +214,19 @@ for var_name in var_names:
                 bedrock = dataset.variables['bedTopography'][0,:]
             else:
                 bedrock = np.zeros(np.shape(thickness))
+            varValue = dataset.variables[var_name][0,:] # get variable array value so we can work from memory and not disk
 
             keepCellMask = np.zeros((nCells,), dtype=np.int8)
 
             if mask_scheme == 'grd':
-                keepCellMask[thickness*ice_density/ocean_density + bedrock > 0.0] = 1
+                keepCellMask[(thickness*ice_density/ocean_density + bedrock > 0.0) * (thickness>0.0)] = 1
             # find the mask for grounded ice region
             elif mask_scheme == 'all':
                 keepCellMask[thickness > 0.0] = 1
 
             keepCellMaskNew = np.copy(keepCellMask)  # make a copy to edit that will be used later
-            keepCellMaskOld = np.copy(keepCellMask)  # make a copy to edit that can be edited without changing the original
+            keepCellMaskOrig = np.copy(keepCellMask)  # make a copy to edit that can be edited without changing the original
+
 
             # recursive extrapolation steps:
             # 1) find cell A with mask = 0
@@ -239,9 +241,11 @@ for var_name in var_names:
 
                 keepCellMask = np.copy(keepCellMaskNew)
                 searchCells = np.where(keepCellMask==0)[0]
+                varValueOld = np.copy(varValue)
 
                 for iCell in searchCells:
                     neighborcellID = cellsOnCell[iCell,:nEdgesOnCell[iCell]]-1
+                    neighborcellID = neighborcellID[neighborcellID>=0] # Important: ignore the phantom "neighbors" that are off the edge of the mesh (0 values in cellsOnCell)
 
                     mask_for_idx = keepCellMask[neighborcellID] # active cell mask
                     mask_nonzero_idx, = np.nonzero(mask_for_idx)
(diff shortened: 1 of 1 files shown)


matthewhoffman:
@trhille , unfortunately, we are going to have to redo this branch onto this version of the code that includes previous unmerged changed:
https://github.com/matthewhoffman/MPAS-Tools/commits/add_exodus_conversion
We can try rebasing it, but I think the conflicts will be too confusing, so it will make more sense to just manually re-create the changes you've made onto this version and update the branch.

matthewhoffman:
@trhille , the branch referenced in my previous message has now been merged into `master` (#219 ), so you should re-build your branch from master.  I tried rebasing your branch just now, and the conflicts that results are too confusing to resolve easily, so the best approach will be to just make a new branch and manually make the required changes.  Have a go at that and let me know if you run into any trouble.  Once that is complete, I can show you how to push the new branch over the top of the one in this PR so the PR does not have to change.

xylar:
@trhille and @matthewhoffman, is this a script you would like to include in the `mpas_tools` conda package?  If so, should I hold off on the next release until this is merged?  See https://github.com/MPAS-Dev/MPAS-Tools/pull/293

xylar:
It doesn't look like this script is currently part of the package:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/conda_package/setup.py#L48-L51
but it would be easy to add if you want.

trhille:
I'll defer to @matthewhoffman on whether to include it in the package. However, we've come up with some ideas about how to effect these changes better, so the merge may be a few days off or more.

matthewhoffman:
@xylar , thanks for checking.  At this point, there is no specific need to have this script be part of mpas_tools conda package.  It is not a part of any automated workflows, and the script is still in fairly regular development, so it makes more sense to leave it out for the time being.

xylar:
Okay, thanks!  I'm going ahead with the release then.

landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
```diff
@@ -1,26 +1,27 @@
...
+#The script will use an environment variable SEACAS to find the Exodus python module.
+#To install the exodus python module, follow instructions here:
+#    https://github.com/gsjaardema/seacas
+#Note that you can follow the shortened instructions under "Exodus", and it is not necessary
+#to install the entire SEACAS set of libraries.
+#
+#Created on Tue Feb 13 23:50:20 2018
+#
+#@author: Tong Zhang, Matt Hoffman, Trevor Hillebrand
+#"""
```
matthewhoffman: Is there a reason you had to comment this block?  If not, can you add a commit restoring it?

landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
```diff
@@ -45,9 +46,9 @@
 SEACAS_path = os.getenv('SEACAS')
 if SEACAS_path == None:
    #sys.path.append('/home/tzhang/Apps/seacas/lib')
-   #sys.path.append('/Users/trevorhillebrand/Documents/mpas/seacas/lib/')
+   sys.path.append('/Users/trevorhillebrand/Documents/mpas/seacas/lib/')
```
matthewhoffman: Can you return the changes in this if-statement back to the original?

landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
```diff
@@ -333,25 +334,55 @@ def interpolateTemperature():
...
-    
-    # Update history attribute of netCDF file
-    thiscommand = datetime.now().strftime("%a %b %d %H:%M:%S %Y") + ": " + " ".join(sys.argv[:])
-    if hasattr(dataset, 'history'):
-       newhist = '\n'.join([thiscommand, getattr(dataset, 'history')])
+    print("\n{} extrapolation and smoothing finished!".format(var_name))
+    return dataset
+ 
+# === Step 4: Call functions to perform conversion, extrapolation, and smoothing =============
+# Define the variable(s) you are converting. Currently supports a single variable or "all"
```
matthewhoffman: Update comment to include support for a comma-delimited list.

landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
```diff
@@ -253,9 +254,9 @@ def interpolateTemperature():
     # 6) go to step 1)
     
     if var_name == 'thickness':
```
matthewhoffman: I'm thinking that we only want to do extrapolation and smoothing for beta, stiffnessFactor, and maybe surfaceAirTemperature, and not for any of the other fields (the other thermal related fields need only be defined where there is ice because in principle they should be evolved with the ice, and even if they are not we don't necessarily just want some simple extrapolation applied.  uReconstructX/Y are model solutions and don't have a need to be defined outside the ice extent) .  What do you think?

landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
```diff
@@ -139,88 +151,77 @@ def get_CellID_array():
...
-    if var_name == "temperature":
-        dataset = interpolateTemperature()
-
+    return dataset
     
-    print("Successful in converting data from Exodus to MPAS!")
 
-    # === Step 2: Optional extrapolation =============
-    
+# === Step 2: Optional extrapolation =============
```
matthewhoffman: I recommend removing the "Step X" from these comments now that they are functions - it's not necessarily the case from the function definition that they will be called in the order listed.

landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
trhille: That was accidental. Uncommenting.

matthewhoffman:
@trhille , also please remove any trailing whitespace.

PR #290, opened 2020-02-24
Add logger to conversion module of conda package

xylar:
This allows the output to be sent to a log file instead of stdout/stderr


xylar:
I tested this on a local build in combination with https://github.com/MPAS-Dev/MPAS-Analysis/pull/680 and it works as expected.

PR #296, opened 2020-03-05
Fix reading of sphere_radius, x_period and mesh_spec

xylar:
https://github.com/MPAS-Dev/MPAS-Tools/pull/279 fixed y_period but other attributes were still being read as strings and converted to doubles (even though they are stored in the NetCDF file as doubles).  The remaining variables need to be fixed in the same way.

### DIFF
--- a/compass/meta.yaml
+++ b/compass/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "compass" %}
-{% set version = "0.1.0" %}
+{% set version = "0.1.1" %}
 
 package:
   name: {{ name|lower }}
@@ -7,7 +7,7 @@ package:
 
 build:
   noarch: python
-  number: 1
+  number: 0
 
 requirements:
   host:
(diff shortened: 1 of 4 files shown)


xylar:
## Testing

I build the conda package with this source.  With the new build, I ran:
```
$ MpasCellCuller.x ocean.QU.480km.151209.nc test.nc
$ ncdump -h test.nc
...
// global attributes:
		:on_a_sphere = "YES" ;
		:sphere_radius = 6371229. ;
```
This is the expected value.

Thanks for the help identifying the problem @jonbob!

xylar:
@vanroekel and @jonbob, if you could glance at the changes and make sure:
1. you're okay with the 3 places I removed `atof()` calls
2. you don't see any that I missed this time
that would be great!

There are also some white-space changes that happen automatically in my editor, so don't worry about those.

I also updated the version number because this fix is worthy of a release.

PR #298, opened 2020-03-07
Add NCO, ESMF, pyremap and mpich to compass env.

xylar:
These are needed by the e3sm_coupling step


xylar:
I ran into an issue where NCO and ESMF were installed in my test environment on my laptop (though why is a bit of a mystery) but the ESMF version was not the one that supports `mpich` and parallel generation of mapping files was failing.  I am not *sure* this will fix it but we surely want the mpich version of ESMF regardless.

PR #310, opened 2020-05-15
Crop to a region in the ParaView extractor

xylar:
If a geojson file (or `geometric_features.FeatureCollection`) is provided to the ParaView VTK extractor, use conversion tools to create a mask, cull the mesh to that mask, and extract the requested fields in the masked region.

The default behavior has been changed so that `all`, `allOnCells`, `allOnVertices`, and `allOnEdges` refer only to variables in the time-series files by default, and will include variables from the mesh file only if `--include_mesh_vars` is specified.

Some shared functionality has been pulled out into helper functions.

### DIFF
--- a/conda_package/mpas_tools/viz/paraview_extractor.py
+++ b/conda_package/mpas_tools/viz/paraview_extractor.py
@@ -465,6 +465,10 @@ def main():
                         'name': 'mask',
                         'object': 'region'}
             for feature in featuresDict['features']:
+                if feature['geometry']['type'] not in ['Polygon',
+                                                       'MultiPolygon']:
+                    raise ValueError('All masking features must be regions '
+                                     '(Polygons or MultiPolygons)')
                 # assign the default values if they're not already present
                 for key, value in defaults.items():
                     if key not in feature['properties']:
--- a/conda_package/mpas_tools/viz/paraview_extractor.py
+++ b/conda_package/mpas_tools/viz/paraview_extractor.py
@@ -85,18 +85,28 @@ from builtins import input
 
 from netCDF4 import Dataset as NetCDFFile
 
+import xarray
+import json
+from geometric_features import FeatureCollection
+import logging
+from io import StringIO
+
 try:
     from progressbar import ProgressBar, Percentage, Bar, ETA
     use_progress_bar = True
 except ImportError:
     use_progress_bar = False
 
+from mpas_tools.conversion import mask, cull
+from mpas_tools.io import write_netcdf
+
 
 def extract_vtk(filename_pattern, variable_list='all', dimension_list=None,
                 mesh_filename=None, blocking=10000, output_32bit=False,
                 combine=False, append=False, out_dir=None, xtime='xtime',
                 lonlat=False, time=None, ignore_time=False, topo_dim=None,
-                topo_cell_index=None):
+                topo_cell_index=None, include_mesh_vars=False,
+                fc_region_mask=None, temp_dir='./culled_region'):
     """
     Extract fields from a time series of NetCDF files as VTK files for plotting
     in paraview.
@@ -233,6 +243,18 @@ def extract_vtk(filename_pattern, variable_list='all', dimension_list=None,
     topo_cell_index: str, optional
         Index array indicating the bottom of the domain (default is the
         topo_dim-1 for all cells)
+
+    include_mesh_vars: bool, optional
+        Whether to include mesh variables as well as time-series variables
+        in the extraction
+
+    fc_region_mask: geometric_features.FeatureCollection, optional
+        A feature collection used to define a mask.  The MPAS data is culled to
+        lie within the mask before conversion to VTK proceeds
+
+    temp_dir: str, optional
+        If fc_region_mask is supplied, a temporary directory where the culled
+        mesh and time series files are stored
     """
 
     if ignore_time:
@@ -249,6 +271,12 @@ def extract_vtk(filename_pattern, variable_list='all', dimension_list=None,
     if mesh_filename is None:
         mesh_filename = time_file_names[0]
         separate_mesh_file = False
+
+    if fc_region_mask is not None:
+        mesh_filename, time_file_names = _cull_files(
+            fc_region_mask, temp_dir, mesh_filename, time_file_names,
+            separate_mesh_file, variable_list, include_mesh_vars, xtime)
+        separate_mesh_file = True
 
     # Setting dimension values:
     time_series_file = open_netcdf(time_file_names[0])
@@ -267,7 +295,7 @@ def extract_vtk(filename_pattern, variable_list='all', dimension_list=None,
     (all_dim_vals, cellVars, vertexVars, edgeVars) = \
         setup_dimension_values_and_sort_vars(
             time_series_file, mesh_file,  variable_list, extra_dims,
-            basic_dims=basic_dims)
+            include_mesh_vars, basic_dims=basic_dims)
     time_series_file.close()
     if mesh_file is not None:
         mesh_file.close()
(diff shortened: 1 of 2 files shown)


xylar:
> The default behavior has been changed so that all, allOnCells, allOnVertices, and allOnEdges refer only to variables in the time-series files by default, and will include variables from the mesh file only if --include_mesh_vars is specified.

@pwolfram and @sbrus89, let me know if this change in default behavior will give you trouble.

xylar:
## Testing

I tested this on 8 years of output from a QU240wLI simulation.  I tested both the time series data and just the restart file on its own.  I generated a geojson file at http://geojson.io and didn't put in the feature name, component or object properties, so these appear to be added by default as expected.  (Less for users to have to worry about).

Here is the region I made:
image
And here is an example in ParaVeiw:
bottom_depth

xylar:
Similar idea but this time with a higher res mesh to show that that seems feasible:
weddell

I don't have a long time series with a high res mesh handy.

xylar:
@pwolfram, this is ready for you to try out and review once you get a chance.  You should be able to just run with `compass_0.1.5` but give the full path to the stub script, e.g.:
```
source /usr/projects/climate/SHARED_CLIMATE/anaconda_envs/base/etc/profile.d/conda.sh
conda activate compass_0.1.5
~/climate/mpas_work/tools/paraview_extractor_region/visualization/paraview_vtk_field_extractor/paraview_vtk_field_extractor.py -f restarts/restart.0001-01-21_00.00.00.nc -v allOnCells --combine --region_mask region.geojson
```
(Change the path to your own check-out of the code in case you need to make changes.)

pwolfram:
Thanks @xylar -- trying this out (using cool http://geojson.io/#map=8/39.021/-74.800 mentioned yesterday for a quick draw without the greater challenge of using GIS).

Confirmed this works using your local install `/users/xylar/`.  This is really great -- works like a charm.  Thanks for bringing in this feature (unexpected but really appreciated)!  Tested this with multiple time steps.  This is *really* nice and a clean solution.

This works really great to pull down ~50 frames in regional, high-res (2 km) area!
image

conda_package/mpas_tools/viz/paraview_extractor.py
```diff
@@ -233,6 +243,18 @@ def extract_vtk(filename_pattern, variable_list='all', dimension_list=None,
     topo_cell_index: str, optional
         Index array indicating the bottom of the domain (default is the
         topo_dim-1 for all cells)
+
+    include_mesh_vars: bool, optional
+        Whether to include mesh variables as well as time-series variables
+        in the extraction
```
pwolfram, line 249: Smart addition-- seems like an obvious thing we already should have been using.  May break some existing code but I think it is worth it because of the improvement.  cc @sbrus89

conda_package/mpas_tools/viz/paraview_extractor.py
```diff
@@ -415,8 +443,36 @@ def main():
...
+            featuresDict = json.load(f)
+            defaults = {'component': 'ocean',
+                        'name': 'mask',
+                        'object': 'region'}
+            for feature in featuresDict['features']:
+                # assign the default values if they're not already present
+                for key, value in defaults.items():
+                    if key not in feature['properties']:
+                        feature['properties'][key] = value
+                fc_region_mask.add_feature(feature)
```
pwolfram, line 472: @xylar, something I didn't check but may be worth checking for here is to make sure this is a polygon, not a polyline or some other not-appropriate feature that wouldn't work to clip the mesh.  Not a big deal (because feeding back geojson file should make script crash), but would be way to harden the code.  I'm happy to see a merge without this but think someone would raise this issue later, which is why I bring it up now for efficiency.

conda_package/mpas_tools/viz/paraview_extractor.py
```diff
@@ -2013,4 +2032,162 @@ def _fix_periodic_vertices_1D(vertices, verticesOnCell, validVertices,
     return tuple(outVertices), verticesOnCell  # }}}
 
 
+def _expand_variable_list(variable_list, time_series_variables,
```
pwolfram, line 2035: Nice

conda_package/mpas_tools/viz/paraview_extractor.py
```diff
@@ -2013,4 +2032,162 @@ def _fix_periodic_vertices_1D(vertices, verticesOnCell, validVertices,
...
+            valid_cell_on_edge.isel(TWO=0),
+            valid_cell_on_edge.isel(TWO=1))
+
+    if use_progress_bar:
+        widgets = ['Cropping time series to region: ', Percentage(), ' ',
+                   Bar(), ' ', ETA()]
+        bar = ProgressBar(widgets=widgets, maxval=len(time_file_names)).start()
+    else:
+        print('Cropping time series to region')
+        bar = None
```
pwolfram, line 2166: I like how the `*.nc` files are kept in a directory to document this work, in case quantitative analysis is needed, e.g, in xarray.

conda_package/mpas_tools/viz/paraview_extractor.py
```diff
@@ -2013,4 +2032,162 @@ def _fix_periodic_vertices_1D(vertices, verticesOnCell, validVertices,
...
+        ds_out = xarray.Dataset()
+        if xtime is None:
+            ds_in = ds_in[variable_names]
+        else:
+            ds_in = ds_in[variable_names + [xtime]]
+            ds_out[xtime] = ds_in[xtime]
+        for var in ds_in.data_vars:
+            for dim in region_masks:
+                if dim in ds_in[var].dims:
+                    ds_out[var] = ds_in[var].where(region_masks[dim], drop=True)
```
pwolfram, line 2182: Nice, glad to see `drop=True` is useful for this!  Had this in mind a long time ago when getting this into xarray.

conda_package/mpas_tools/viz/paraview_extractor.py
```diff
@@ -2013,4 +2032,162 @@ def _fix_periodic_vertices_1D(vertices, verticesOnCell, validVertices,
...
+        'cellsOnVertex', 'edgesOnVertex', 'indexToVertexID',
+        'kiteAreasOnVertex', 'latVertex', 'lonVertex', 'xVertex', 'yVertex',
+        'zVertex', 'weightsOnEdge']
+
+    try:
+        os.makedirs(temp_dir)
+    except OSError:
+        pass
+
+    log_stream = StringIO()
```
pwolfram, line 2111: @xylar, is this just to help control output?  It wasn't obvious to me if this is saved to disk or not, but presumably this is available for debugging purposes.

pwolfram: Not an issue but would be curious to learn more if you have time, otherwise I can dig into it later as needed.

conda_package/mpas_tools/viz/paraview_extractor.py
```diff
@@ -85,18 +85,28 @@
...
+import logging
+from io import StringIO
+
 try:
     from progressbar import ProgressBar, Percentage, Bar, ETA
     use_progress_bar = True
 except ImportError:
     use_progress_bar = False
 
+from mpas_tools.conversion import mask, cull
```
pwolfram, line 100: Glad to see the broader pythonization is coming into use here!

conda_package/mpas_tools/viz/paraview_extractor.py
xylar: Sure, that's very easy to add.  I though about it, too, but decided to prioritize making it work over making it robust.  I'll push a commit for this in just a minute.

conda_package/mpas_tools/viz/paraview_extractor.py
xylar: That's a side effect of me not wanting to make and delete a temp directory but I agree it's a good thing to keep around.  Theoretically, we would only create missing files (perhaps optionally) but I think it's safest to just overwrite the files if the user wants new analysis.  The circumstances where you run the extractor multiple times and would want to save time by not overwriting files seem pretty unusual.

conda_package/mpas_tools/viz/paraview_extractor.py
xylar: Yep, for sure!

conda_package/mpas_tools/viz/paraview_extractor.py
xylar: This is actually just the simplest way I could figure out to throw away the output.  If you don't pass a logger, it goes to strout/stderr so it turns out to be easiest to pass a logger and just not look at the output.

xylar:
@pwolfram, could you check if the commit I just pushed catches a bad mask file as expected?  I don't have my external drive handy right now and would need it for the data I was testing on.  Can test later if you aren't in a position to.

xylar:
Okay, I tested the last commit, both with a valid and an invalid mask and it worked as expected (raising an error only when I supplied a transect instead of a region).

xylar:
@pwolfram, thanks for the review!  Let me know how this works out for you.  It will be part of the next compass once I get one set up (which I expect to be soon).

pwolfram:
Thanks @xylar, this is really great!  Will let you know if I run into any geojson issues.  Thanks!

PR #312, opened 2020-05-19
Update version of pyremap in compass metapackage

xylar:
The new `pyremap` will detect if `esmf` was installed with MPI or not and will either call `ESMF_RegridWeightGen` with `mpirun` or not as appropriate.  See https://github.com/MPAS-Dev/pyremap/pull/16

Update conda package to v0.1.6

### DIFF
--- a/compass/meta.yaml
+++ b/compass/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "compass" %}
-{% set version = "0.1.5" %}
+{% set version = "0.1.6" %}
 {% set build = 0 %}
 
 {% if mpi == "nompi" %}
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I ran the MPAS-Ocean nightly and land-ice-fluxes regression suites from COMPASS on Grizzly with a test version of the `compass_0.1.6` environment. I also ran the `e3sm_coupling` steps for the QU240 and QU240wISC test cases, which had been failing on 1 core previously (because `ESMF_RegridWeightGen` was getting called without `mpirun`). Everything ran successfully.

xylar:
This new version has just been uploaded to the `e3sm` anaconda channel.  I will deploy it now on grizzly, cori, compy, anvil and cooley.

PR #313, opened 2020-05-28
Update the script for setting up compass environments

xylar:
Update python and compass versions.

Make mpich the default if a machine doesn't override it, but serial is the default on cori, compy and cooley, where I'm not sure mpich works.


xylar:
This got missed in #312

PR #311, opened 2020-05-18
Add mesh creation module to  mpas_tools conda package

xylar:
The major change in this merge is moving `jigsaw_to_MPAS` from [COMPASS](https://github.com/MPAS-Dev/MPAS-Model/tree/ocean/develop/testing_and_setup/compass/ocean/jigsaw_to_MPAS) to the `mpas_tools` conda package.  It is now in `mpas_tools.mesh.creation`.

Other changes in this PR include:
* Moving color map related functionality to the `viz` module
* Moving `mpas_tools.conversion` to `mpas_tools.mesh.conversion` (but `mpas_tools.conversion` still exists for backward compatibility
* Making entry points for relevant tools from `jigsaw_to_MPAS`
* Adding a bit of code to make sure `define_base_mesh.py` gets found in the current directory within the `build_mesh` module
* Update `mpas_tools` to v0.0.11 and `compass` metapackage to v0.1.9
* Add some tests to the recipe to make sure the new entry points and module are present
* Add some documentation of how developers should modify and test the `mpas_tools` package

`geometric_features` in `compass` has been updated to v0.1.11.

PR #314, opened 2020-05-28
Reorganize mesh creation

xylar:
This PR builds on #311 by performing some reorganization of `mpas_tools.mesh.creation`.

It includes:
* Moving `coastal_tools`, `inject_bathymetry`, `inject_meshDensity` and `inject_preserve_floodplain` to `mpas_tools.ocean`, since these are ocean-specific tools.
* `inject_meshDensity` had been broken into 3 functions, one for reading from a file and one each for taking numpy arrays that define spherical or planar meshes.
* Splits out tools for computing signed-distance tools from `coastal_tools`, since the former are general to any component.  These tools are now in `mpas_tools.mesh.creation.signed_distance`
* Add `earth_radius` parameter to jigsaw_driver and a few other places so the Earth radius isn't hard coded with different values in different places
* Break `build_mesh` into polar and spherical functions and move ocean-specific steps to a new module `mpas_tools.ocean.build_mesh`.


xylar:
@pwolfram, @sbrus89, @dengwirda, @mark-petersen, and @proteanplanet, this is just a heads-up that I'm working on reorganizing the mesh creation tools to separate out ocean-specific pieces from generic ones that might be useful for other components.  @dengwirda, you may want to take a look at this and, again, make sure it doesn't conflict with changes you'd like to make.

When things have progressed a big further and if there aren't major protests, I'll make a corresponding PR in MPAS-Model/ocean/develop to make the corresponding changes in existing COMPASS test cases.

xylar:
@mark-petersen and @sbrus89, please always approve the associated MPAS-Tools changes like this PR along with the MPAS-Model PR.  Again, I am reluctantly merging without your review because it was tacitly given in https://github.com/MPAS-Dev/MPAS-Model/pull/577

PR #315, opened 2020-05-28
Add CIME Earth radius

xylar:
This is stored in a dictionary of potential CIME constants.

A test has also been added to verify each time the `mpas_tools` package is created (e.g. in Travis CI or on `conda-forge`) that the constants have the expected value by comparing them with the constants from CIME from `E3SM/master`.

### DIFF
--- a/conda_package/mpas_tools/cime/constants.py
+++ b/conda_package/mpas_tools/cime/constants.py
@@ -1,0 +1,2 @@
+constants = \
+    {'SHR_CONST_REARTH': 6.37122e6}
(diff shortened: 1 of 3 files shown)


xylar:
@proteanplanet, @dengwirda, let me know if this solution works for you.  You could use the CIME Earth radius in your python code via:
```
from mpas_tools.cime.constants import constants

earth_radius = constants['SHR_CONST_REARTH']
```

Also, I can easily add other constants from [CIME](https://raw.githubusercontent.com/E3SM-Project/E3SM/master/cime/src/share/util/shr_const_mod.F90) but I didn't know which would be useful during mesh creation (as opposed to within cores in MPAS-Model).

proteanplanet:
@xylar This looks close to perfect to me - the exact solution we need.

xylar:
Thanks @proteanplanet for the quick review!

dengwirda:
Thanks @xylar this looks great!

xylar:
Thanks @dengwirda!

PR #318, opened 2020-06-10
Strip NULL (\x00) characters from xtime

### LINKED ISSUE #317: paraview_vtk_field_extractor produces extra characters after xtime
I’m using e3sm_unified_1.3.0 on cori, paraview_vtk_field_extractor.py. In the file
`vtk_output/time_series/fieldsOnCells.0.vtp`
line 5 is:
```
      4 <xtime>
      5 0001-01-11_00:00:00^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@
      6 </xtime>
```
The `^@` are some crazy characters vim doesn’t understand. I can delete those by hand, or use   `-x none`. But we want to avoid the creation of the `^@` characters. 

This results in paraview producing an error about invalid xml file format, and not able to open the file.

xylar:
Some NetCDF file types seem to include these.

closes #317

### DIFF
--- a/conda_package/mpas_tools/viz/paraview_extractor.py
+++ b/conda_package/mpas_tools/viz/paraview_extractor.py
@@ -648,8 +648,9 @@ def build_field_time_series(local_time_indices, file_names, mesh_file,
                                                      time_series_file))
                     var = time_series_file.variables[xtimeName]
                     if len(var.shape) == 2:
-                        xtime = var[local_time_indices[time_index],
-                                    :].tostring().decode('utf-8').strip()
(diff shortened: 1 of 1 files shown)


mark-petersen:
@xylar, how did you know is was a null string `\x00`? All I saw in vim was `^@`. Is there an editor that tells you the escape character, or did you evaluate it in python?

xylar:
I printed out the strings in python (which is more explicit about giving hex values for unusual characters).

PR #326, opened 2020-07-21
Fix paraview extractor and macros to be compatible with Paraview 5.8.0

### LINKED ISSUE #325: VTK files generated by paraview_vtk_field_extractor will not load in Paraview version 5.8.0
When using paraview_vtk_field_extractor.py  to generate VTK files from MPAS files, an error occurs when trying to load the resulting VTK files in Paraview version 5.8.0:
`Cannot read cell data from Polys. Offsets and connectivity arrays must be the same type.`
The same file loads without incident in Paraview version 5.6.0 or 5.4.1.

### LINKED ISSUE #327: date annotation macro no longer works in Paraview 5.8.0
First, it seems like Paraview isn't picking up the "timestep" attribute correctly (it starts at 0 even when the time is 500).  Second, there is an issue with the new name of the data structure expected by the annotation filter.

xylar:
First, makes sure that connectivity is a 64-bit integer (was 32-bit because this is how it is stored in MPAS meshes).

Second, fixes the text in the "active" field for cell and point data.

Also fixes the date annotation macro, since this is also a Paraview 5.8.0 issue.

closes #325 
closes #327

### DIFF
--- a/conda_package/mpas_tools/viz/paraview_extractor.py
+++ b/conda_package/mpas_tools/viz/paraview_extractor.py
@@ -1565,7 +1565,7 @@ def build_cell_geom_lists(nc_file, output_32bit, lonlat):  # {{{
                                                           nc_file.x_period,
                                                           nc_file.y_period)
 
-    connectivity = verticesOnCell[validVertices]
+    connectivity = numpy.array(verticesOnCell[validVertices], int)
     offsets = numpy.cumsum(nEdgesOnCell, dtype=int)
     valid_mask = numpy.ones(nCells, bool)
 
--- a/conda_package/mpas_tools/viz/paraview_extractor.py
+++ b/conda_package/mpas_tools/viz/paraview_extractor.py
@@ -1328,8 +1328,7 @@ def write_vtp_header(path, prefix, active_var_index, var_indices,
 
     if cellData:
         vtkFile.openData(str("Cell"),
-                         scalars=[str(var) for var in
-                                  variable_list[active_var_index]])
+                         scalars=variable_list[active_var_index])
         for iVar in var_indices:
             var_name = variable_list[iVar]
             (out_var_names, dim_list) = \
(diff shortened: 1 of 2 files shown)


xylar:
@matthewhoffman, I have not tested this in Paraview 5.8.0 yet but I will shortly.  However, I am quite confident that the problem should now be fixed.  Please ping Alex if you have his GitHub user name.

xylar:
I was able to load without a problem in Paraveiw 5.8.0.  Let me know how it goes for you.

matthewhoffman:
@xylar , thanks for such a rapid fix on this issue!  I will coordinate with Alex to get it tested.

matthewhoffman:
@xylar , is this something that can be tested without rebuilding the whole conda env?  I tried asking Alex to test with just running paraview_extractor.py directly, but that didn't work.    If not, we might trust it's good and let you go ahead and merge it based on your testing.

xylar:
@matthewhoffman, yes, I tested without building the conda environment.  you have to provide the path to the paraview extractor in this branch.  You would need to be in a conda environment with the appropriate package for the latest `mpas_tools`.  The latest `e3sm-unified` is probably the easiest but the latest `compass` should also do I think.

xylar:
To clarify, you would call the stub script that his here:
https://github.com/xylar/MPAS-Tools/blob/fix_paraview_extractor/visualization/paraview_vtk_field_extractor/paraview_vtk_field_extractor.py

matthewhoffman:
I'm confused - by calling the stub script, wouldn't that invoke the version of the relevant files that are compiled into the conda env (`mpas_tools` or `compass`)?  How does one force the stub script to use the path to the locally checked out version?

xylar:
There's a symlink to `mpas_tools` from the branch in the folder with the stub.  That takes priority over the the version of the package in the python environment.

PR #328, opened 2020-07-21
exo to nc conversion: fix beta extrapolation

matthewhoffman:
The old method for extrapolating beta used the grounded ice extent as
the source region for extrapolation.  However, Albany optimization
includes one extended cell beyond that, which was getting clobbered by
the extrapolation step.  This commit fixes that by defining the source
region as the cells that have beta>0 in the exodus file (i.e., keep any
locations where the optimization returned a beta value and extrapolate
out from there.)

This commit also disables the smoothing of the beta extrapolation, which
seems to provide hit or miss results.

### DIFF
--- a/landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
+++ b/landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
@@ -52,7 +52,7 @@ from exodus import exodus
 # Map and copy Exodus data to MPAS data
 
 # Create dictionary of variables that are supported by the script
-mpas_exodus_var_dic = {"beta":"basal_friction", "thickness":"ice_thickness",\
+mpas_exodus_var_dic = {"beta":"basal_friction_log", "thickness":"ice_thickness",\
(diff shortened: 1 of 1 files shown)


matthewhoffman:
@mperego , let me know if you want to test this fix before I merge it.
cc: @trhille

matthewhoffman:
@mperego , ok, I'll make that name change too.  So to be clear, the value of the variable has not changed, just the name?  I.e., the existing conversion line:
```
dataset.variables[var_name][0,cellID_array-1] = np.exp(data_exo_layer) * 1000.0
```
does not need to be changed?

mperego:
> @mperego , ok, I'll make that name change too. So to be clear, the value of the variable has not changed, just the name? I.e., the existing conversion line:
> 
> ```
> dataset.variables[var_name][0,cellID_array-1] = np.exp(data_exo_layer) * 1000.0
> ```
> 
> does not need to be changed?

Yes, that's correct!  B.t.w. I tested it and it seems to work fine.

PR #341, opened 2020-09-01
Update Earth radius in build_mesh to the CIME value

xylar:
Partly addresses https://github.com/MPAS-Dev/MPAS-Model/issues/549


xylar:
@proteanplanet, @dengwirda, if you could take a look just at the last commit of this PR and let me know if it looks correct to you, that would be helpful.  The rest is from #314 and will disappear once that gets merged and I can rebase.

xylar:
> Just to double-check, is the idea to use `../mpas_tools/tests/test_cime_constants.py` to ensure this stays in-sync with `../E3SM-project/E3SM/cime/`?

Yes, If you look at tests, which get run every time there's a PR, a merge to `master` or a new tag, we check that things match E3SM's version of CIME:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/conda_package/mpas_tools/tests/test_cime_constants.py#L5

The same would be true if we add other constants from CIME in the future.

xylar:
Oh, shoot!  Except, I now see that test is broken but it's not giving an error.  Work to do...

xylar:
Implicitly approved by @mark-petersen and @sbrus89 via https://github.com/MPAS-Dev/MPAS-Model/pull/680

proteanplanet:
Sorry @xylar, it wasn't clear whether you'd finished making your changes.  So I held off.

xylar:
@proteanplanet, sorry about that.  I will try to make that clearer in the future by keeping my PRs either as drafts or as "in progress" until they are ready to review and pinging you when the time comes.

I am moving on with this one pretty fast because it is holding up E3SM v2 critical work.

PR #343, opened 2020-09-02
Fix issue with bounds on coastal boxes

xylar:
It is not very safe to use a non-integer step in `arange` because round-off causes trouble.  Also, in this case you want to include
the end point in the `arange` and this was not happening with the previous code.  As a result, global boxes did not include +180 longitude and +90 latidude.

This merge also raises an exception if `mesh_type` is not one of the known values, so the error message is clearer in this case.


xylar:
This is based off of #314 because it modifies files that were moved in that PR.

xylar:
## Testing

With this branch, I get a base mesh in the `coastal/USDEQU120cr10rr2/build_mesh` test case that is 3601 x 1801, as expected, instead of 3600 x 1800.  The min and max lon and lat are -180, 180, -90 and 90, as expected.

With this change, interpolation of mesh density will no longer require adding of an extra element to make the mesh periodic and to include the north pole.

xylar:
Implicitly approved by @mark-petersen and @sbrus89 via https://github.com/MPAS-Dev/MPAS-Model/pull/680

PR #344, opened 2020-09-02
Use an internal bilinear interpolation instead of scipy for meshDensity and bathymetry

xylar:
The `scipy` interpolators are quite slow and inefficient.  For bilinear interpolation on a regular grid, it is much more efficient to use `numpy.interp` in 1D to set up coefficients for doing 2D, bilinear interpolation.  A new function `mpas_tools.mesh.interpolation.interp_bilin()` is added to do this.


xylar:
This work is based off of #314 and #343.

xylar:
## Testing

I tested this with the `coastal/USDEQU120cr10rr2/build_mesh/` test case and the resulting `meshDensity` field looks similar to what was obtained previously.  I need to do a proper `diff` and make sure changes are at machine roundoff for both `meshDensity` and `bathymetry`.  

Interpolation is **significantly** faster with the new routine.  I didn't time the difference but it is noticeable.

xylar:
Implicitly approved by @mark-petersen and @sbrus89 via MPAS-Dev/MPAS-Model#680

PR #346, opened 2020-09-04
Allow conversion tools to be called with a directory 

xylar:
The directory is used as the base path for creating a temporary directory for files produced during mesh conversion, cell culling or mask creation.

The default behavior of using `/tmp` is causing trouble in MPAS-Analysis, see https://github.com/MPAS-Dev/MPAS-Analysis/issues/729

### DIFF
--- a/compass/meta.yaml
+++ b/compass/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "compass" %}
-{% set version = "0.1.10" %}
+{% set version = "0.1.11" %}
 {% set build = 0 %}
 
 {% if mpi == "nompi" %}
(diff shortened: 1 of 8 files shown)


xylar:
## Testing

I tested this both by running `QU240` and `QU240wISC` test cases and by running MPAS-Analysis on a `QU240wLI` run where I had moved the cached masks and mapping files to a temporary directory to force them to be re-created.  Relative paths seemed to cause some trouble but converting them to absolute paths seems to have solved that.

xylar:
This fix is needed for an MPAS-Analysis release that is in turn needed for an E3SM-Unified update, so I will merge as soon as testing finished successfully.

PR #349, opened 2020-09-22
A lot of new additions to the documentation

xylar:
This merge adds documentation for all of the remaining functionality in the `mpas_tools` conda package with the exception of the scripts copied from the `landice` directory.

This merge also fixes some very minor bugs related to a deprecated xarray method (`drop` --> `drop_vars`) and some warnings about unneeded parentheses that were discovered in the process of improving docstrings.


conda_package/docs/ocean/coastal_tools.rst
```diff
@@ -0,0 +1,354 @@
+.. _ocean_coastal_tools:
+
+.. |---| unicode:: U+2014  .. em dash, trimming surrounding whitespace
+   :trim:
+
+Coastal Tools
+=============
```
xylar, line 7: @sbrus89, could you look over this documentation and make sure it is correct?  I tried my best to document these functions (including adding detailed docstrings) but there wasn't a whole lot of existing comments to work with so I had to reverse engineer in a few cases.

conda_package/mpas_tools/ocean/coastal_tools.py
```diff
@@ -1,11 +1,11 @@
 #!/usr/bin/env python
-'''
+"""
 name: coastal_tools
 authors: Steven Brus
 
 last modified: 07/09/2018
```
xylar, line 7: @sbrus89, could you look at the docstrings I have added to this module to make sure they are correct?

conda_package/mpas_tools/ocean/coastal_tools.py
```diff
@@ -561,8 +739,77 @@ def distance_to_coast(coastlines, lon_grd, lat_grd, origin, nn_search, smooth_wi
...
+        The extent of the plot if ``plot_option=True``
+
+    lon_grd : ndarray
+        A 1D array of longitudes in degrees in the range from -180 to 180 used
+        in plotting if ``plot_option=True`` (the same as ``lon``)
+
+    lat_grd : ndarray
+        A 1D array of latitudes in degrees in the range from -90 to 90 used
+        in plotting if ``plot_option=True`` (the same as ``lat``)
+
```
xylar, line 798: @sbrus89, reading the code, I don't see how `lon_grd` and `lat_grd` can be different from `lon` and `lat`.  They both have to be the coordinates of `cell_width` because `lon` and `lat` are used to determine the indices into `cell_width` for a given quad, but then `lon_grd` and `lat_grd` are use with `cell_width` in a call to `get_data_inside_box()`, which suggests to me that those also have to be the coordinates for `cell_width`.  If these are redundant, maybe we can do a PR to remove them?

conda_package/mpas_tools/ocean/coastal_tools.py
```diff
@@ -479,8 +610,55 @@ def extract_coastlines(nc_file, nc_vars, region_box, z_contour=0, n_longest=10,
...
+    lon_grd : ndarray
+        A 1D array of longitudes in degrees in the range from -180 to 180
+
+    lat_grd : ndarray
+        A 1D array of latitudes in degrees in the range from -90 to 90
+
+    origin : ndarray
+        A lon/lat point defining the origin, not currently used by the code
+
```
xylar, line 633: @sbrus89, `pycharm` tells me that this argument isn't used (anymore?) so could we remove it?  Or is it being kept for
backwards compatibility?

xylar:
@sbrus89, if you could check over the documentation I added for code you wrote, that would make me feel more secure about it.

@dengwirda and @proteanplanet, your comments on this documentation now or at a later point are welcome but not needed.  Hopefully, it will help you with getting more involved in MPAS-Tools.

@mark-petersen, just giving you a chance to chime in if you want -- a few of the scripts I documented here are yours -- but I think I know the functionality well enough that the documentation is accurate.

xylar:
If anyone would like to do a local test build of the documentation, instructions are here:
http://mpas-dev.github.io/MPAS-Tools/stable/building_docs.html

Here is the script I use to build the conda package and then the documentation.
``` bash
#!/bin/bash

source /home/xylar/miniconda3/etc/profile.d/conda.sh

# exit if a subprocess ends in errors
set -e

PYTHON_VERSION=3.8

rm -rf ~/miniconda3/conda-bld

conda build -m "conda_package/ci/linux_python${PYTHON_VERSION}.yaml" "conda_package/recipe"

conda create --yes --quiet --name test_mpas_tools_docs --use-local python=$PYTHON_VERSION \
        mpas_tools sphinx mock sphinx_rtd_theme m2r
conda activate test_mpas_tools_docs

DOCS_VERSION=`python -c "import mpas_tools; print(mpas_tools.__version__)"`
export DOCS_VERSION
echo "Docs version: $DOCS_VERSION"

cd conda_package/docs || exit 1
make html
```

Then, you just need to open `_build/html/index.html` in your browser.

conda_package/mpas_tools/planar_hex.py
```diff
@@ -64,9 +64,9 @@ def make_planar_hex_mesh(nx, ny, dc, nonperiodic_x,
     compute_coordinates(mesh)
     add_one_to_indices(mesh)
 
-    # drop some arrays that aren't stantard for MPAS but were used to compute
+    # drop some arrays that aren't standard for MPAS but were used to compute
     # the hex mesh
-    mesh = mesh.drop(['cellIdx', 'cellRow', 'cellCol'])
+    mesh = mesh.drop_vars(['cellIdx', 'cellRow', 'cellCol'])
```
xylar, line 69: This is the deprecated method fix.

conda_package/mpas_tools/planar_hex.py
```diff
@@ -127,14 +127,14 @@ def initial_setup(nx, ny, dc, nonperiodic_x, nonperiodic_y):
                                       numpy.arange(ny, dtype='i4'))
 
     mesh['cellIdx'] = (('ny', 'nx'), cellIdx)
-    mesh['cellRow'] = (('nCells'), cellRow.ravel())
-    mesh['cellCol'] = (('nCells'), cellCol.ravel())
+    mesh['cellRow'] = (('nCells',), cellRow.ravel())
```
xylar, line 130: Changes like this one are to fix a warning about redundant parentheses.  By adding a comma, these become `tuples`, which is what I had originally intended.

conda_package/docs/ocean/coastal_tools.rst
sbrus89: @xylar, of course. I will look this over. Thanks very much for cleaning this up!

conda_package/mpas_tools/ocean/coastal_tools.py
sbrus89: @xylar, I don't believe this is used anymore. We should remove it.

conda_package/docs/ocean/coastal_tools.rst
sbrus89: One thing we should discuss is making the `coastal_tools.py` functions work with geometric features instead of the specifying boxes. Right now it is very clumsy.

conda_package/mpas_tools/ocean/coastal_tools.py
sbrus89: I'm sure you're right, let me check on this.

conda_package/mpas_tools/ocean/coastal_tools.py
sbrus89: I'm sure you are right, let me check on this.

conda_package/docs/ocean/coastal_tools.rst
xylar: Yes, I think that would, indeed, be a better approach.  As time allows...

mark-petersen:
@xylar, this is great! Thanks for pointing it out to me.

xylar:
@sbrus89, could you give me an estimate on when you might be able to review/approve this PR?

sbrus89:
@xylar, I'm working on this today.

conda_package/docs/ocean/coastal_tools.rst
```diff
@@ -0,0 +1,354 @@
...
+.. _coastal_tools_regions:
+
+Regions
+-------
+
+:ref:`coastal_tools_extract` requires a set of bounding regions to be defined.
+These regions are made up of a list of quadrilaterals to ``include`` and another
+list to ``exclude``.  The quadrilaterals are either bounding rectangles
+(min lon, max lon, min lat, max lat) or lists of 4 (lon, lat) points.
+
```
sbrus89, line 309: It may be helpful to say something like:

``include`` boxes specify areas where coastline contours will be extracted from the underlying bathymetry dataset. ``exclude`` boxes can be used to select regions within ``include`` regions where coastline extraction should not be done. For example, a large ``include`` box covering the U.S. East Coast may contain small islands in the Atlantic which may need to be ignored when placing coastal refinement. In this case, ``exclude`` boxes can be specified to eliminate the small coastlines.

conda_package/docs/ocean/coastal_tools.rst
```diff
@@ -0,0 +1,354 @@
...
+The most important parameters to set are ``params['dx_min_coastal']``, the
+resolution in meters of the refined region; ``params['trans_start']``, the
+distance from the coast in meters where the transition in resolution should
+start; and ``params['trans_width']``, the width of the transition itself in
+meters.
+
+The resolution refinement can be confined to a region using
+``params['restrict_box']`` to supply a region of made up of quadrilaterals to
+``include`` and ``exclude`` from the restricted region.  For example, a
```
sbrus89, line 283: Maybe add something like this here:

``include`` boxes specify regions where the distance-based coastal refinement function should be used to update the desired cell width values. ``exclude`` boxes eliminate areas within the ``include`` regions so they are not affected by the coastal refinement function. This is useful for preventing resolution from appearing on one side of a narrow piece of land. For example, coastal refinement in the Gulf of Mexico should not appear on the other Pacific Ocean side of Central America. The ``params['restrict_box']`` regions can be used to enforce this.

conda_package/docs/ocean/coastal_tools.rst
```diff
@@ -0,0 +1,354 @@
...
+* ``'point_list'`` - An optional list of points to add to the coastline
+* ``'plot_option'`` - Whether to plot the extracted coastline.
+* ``'plot_box'`` - If ``plot_option`` is ``True``, the bounds of the plot in
+  longitude and latitude.
+
+Next, the distance to the coastal contours is computed using
+:py:func:`mpas_tools.ocean.coastal_tools.distance_to_coast()` with the
+following values from ``params``:
+
+* ``'origin'`` - A lon and lat point---no longer used in the code
```
sbrus89, line 119: You are right that this should be removed.

conda_package/docs/ocean/coastal_tools.rst
```diff
@@ -0,0 +1,354 @@
...
+
+.. _coastal_tools_regions:
+
+Regions
+-------
+
+:ref:`coastal_tools_extract` requires a set of bounding regions to be defined.
+These regions are made up of a list of quadrilaterals to ``include`` and another
+list to ``exclude``.  The quadrilaterals are either bounding rectangles
+(min lon, max lon, min lat, max lat) or lists of 4 (lon, lat) points.
```
sbrus89, line 308: lists of 4 (lon, lat) points numbered counter clockwise.

conda_package/mpas_tools/ocean/coastal_tools.py
sbrus89: I agree `lon_grd` and `lat_grd` are redundant and should be replace with `lon` and `lat` in this function.

sbrus89 (commented):
@xylar, this looks great. I have left a few comments regarding being more specific about the roles of the `include` and exclude` boxes for the separate purposes of coastline extraction and resolution specification.

conda_package/docs/ocean/coastal_tools.rst
xylar: Okay, I'll do a separate PR for cleaning up these arguments.

conda_package/mpas_tools/ocean/coastal_tools.py
xylar: Okay, I'll do a separate PR for cleaning up these arguments.

xylar:
@sbrus89, I made the changes you requested.  Please let me know if there is anything else that you notice.

sbrus89 (approved):
@xylar, thanks for including those changes. This looks ready to go. Thanks again for your work on this !

xylar:
Thanks for the review, @sbrus89!

PR #351, opened 2020-09-23
Switch mesh_spec attribute from double to string in MPAS cell culler

### LINKED ISSUE #350: MPAS cell culler outputs incorrect mesh_spec
The MPAS cell culler writes out the `mesh_spec` attribute as a double instead of a string.  It converts the ascii character `1` into its numerical value (49) and exports this as a double.  Then, subsequent runs of `ocean_model` (or presumably other MPAS components convert this to `0.0`, perhaps because it is not recognized.

The mask creator and mesh converter correctly specify `mesh_spec` as a string, and the code from these tools should be copied into cell culler.

This is the cause of https://github.com/MPAS-Dev/MPAS-Model/issues/701

xylar:
The `mesh_spec` is stored as a string in other tools, and this is what the MPAS components expect.  Only the cell culler and the supporting utilities had it (incorrectly) as a double, and these are fixed in this merge.

closed #350

### DIFF
--- a/mesh_tools/mesh_conversion_tools/mpas_cell_culler.cpp
+++ b/mesh_tools/mesh_conversion_tools/mpas_cell_culler.cpp
@@ -25,7 +25,7 @@ double sphere_radius, xPeriod, yPeriod;
 string in_history = "";
 string in_file_id = "";
 string in_parent_id = "";
-double in_mesh_spec = 1.0;
+string in_mesh_spec = "1.0";
 bool outputMap = false;
 
 // Connectivity and location information {{{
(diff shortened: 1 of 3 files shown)


xylar:
@mark-petersen, I am currently testing this on the QU240 workflow, which previously produced an incorrect `mesh_spec` during cell culling (49.0 as a double), which was then replaced with another incorrect value during `initial_state` (0.0 as a string).

I think it would be sufficient for you to review this by inspection for now, but I will set up a test `compass` environment before the next release of `mpas_tools` and we can double check that way that this issue is fixed before we make the next release.  I just don't want to set up a test environment for every PR...

xylar:
## Testing

I ran the `ocean/global_ocean/QU240/init` test case from COMPASS on my laptop (Ubuntu 18.04).  I ran the `culled_mesh` step with a test environment created from this branch, and `mesh_spec = '1.0'` in `culled_mesh.nc`, as expected.  I then ran `initial_state` and `mesh_spec = '1.0'` in `initial_state.nc` as well.  So I think this is now fixed.

mark-petersen (approved):
Yes, this looks good to me. Your testing with this is sufficient, thanks.

PR #352, opened 2020-09-26
Fix grid_imask in scrip_from_mpas

xylar:
Previously, the code for computing `grid_imask` was erroneously within an if statement that did not get called for cells with the maximum number of vertices.

The code for filling in vertices has also been rearranged to loop over vertices on a cell (typically 7) rather than cells (typically tens of thousands to millions) for efficiency and to set a better example for future codes that may be build off of this one.

### DIFF
--- a/conda_package/mpas_tools/scrip/from_mpas.py
+++ b/conda_package/mpas_tools/scrip/from_mpas.py
@@ -39,7 +39,7 @@ def scrip_from_mpas(mpasFile, scripFile, useLandIceMask=False):
     lonCell = fin.variables['lonCell'][:]
     latVertex = fin.variables['latVertex'][:]
     lonVertex = fin.variables['lonVertex'][:]
-    verticesOnCell = fin.variables['verticesOnCell'][:]
+    verticesOnCell = fin.variables['verticesOnCell'][:] - 1
     nEdgesOnCell = fin.variables['nEdgesOnCell'][:]
     nCells = len(fin.dimensions['nCells'])
     maxVertices = len(fin.dimensions['maxEdges'])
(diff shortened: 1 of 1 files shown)


xylar:
The main bug this PR fixes was introduced 4 years ago in this commit: https://github.com/MPAS-Dev/MPAS-Tools/commit/5f2976b04e07c2e17d364c7d0964a4b99e478989

The logic for filling in `grid_imask` definitely doesn't belong in the if statement it's in (the bug) but it also doesn't need to be in a loop over cells.

conda_package/mpas_tools/scrip/from_mpas.py
```diff
@@ -94,30 +94,28 @@ def scrip_from_mpas(mpasFile, scripFile, useLandIceMask=False):
...
-                latVertex[verticesOnCell[iCell, vertexMax-1] - 1]
-            grid_corner_lon_local[iCell, vertexMax:] = \
-                lonVertex[verticesOnCell[iCell, vertexMax-1] - 1]
-
-            if useLandIceMask:
-                # If useLandIceMask are enabled, mask out ocean under land ice.
-                grid_imask[iCell] = 1 - landIceMask[0, iCell]
-            else:
-                # If landiceMasks are not enabled, don't mask anything out.
-                grid_imask[iCell] = 1
```
xylar, line 120: The bug is that any cell where `nEdgesOnCell == maxVertices` will not have `grid_imask` set because of this `if` statement.

conda_package/mpas_tools/scrip/from_mpas.py
```diff
@@ -94,30 +94,28 @@ def scrip_from_mpas(mpasFile, scripFile, useLandIceMask=False):
...
+        mask = iVertex >= nEdgesOnCell
+        grid_corner_lat_local[mask, iVertex] = latVertex[lastValidVertex[mask]]
+        grid_corner_lon_local[mask, iVertex] = lonVertex[lastValidVertex[mask]]
+
+    if useLandIceMask:
+        # If useLandIceMask are enabled, mask out ocean under land ice.
+        grid_imask[:] = 1 - landIceMask[0, :]
+    else:
+        # If landiceMasks are not enabled, don't mask anything out.
+        grid_imask[:] = 1
```
xylar, line 117: There is no reason to even have creation of `grid_imask` in a cell loop at all.

conda_package/mpas_tools/scrip/from_mpas.py
```diff
@@ -94,30 +94,28 @@ def scrip_from_mpas(mpasFile, scripFile, useLandIceMask=False):
...
+    for iVertex in range(maxVertices):
+        mask = iVertex < nEdgesOnCell
+        grid_corner_lat_local[mask, iVertex] = \
+            latVertex[verticesOnCell[mask, iVertex]]
+        grid_corner_lon_local[mask, iVertex] = \
+            lonVertex[verticesOnCell[mask, iVertex]]
+
+        mask = iVertex >= nEdgesOnCell
+        grid_corner_lat_local[mask, iVertex] = latVertex[lastValidVertex[mask]]
+        grid_corner_lon_local[mask, iVertex] = lonVertex[lastValidVertex[mask]]
```
xylar, line 110: Looping over vertices on a cell instead of cells is going to be **way** more efficient.

xylar:
## Testing

I ran the QU240 and QU240wISC test cases on my laptop.  I ran the scrip-file generation with the last release of MPAS-Tools (from 3 weeks ago) and again with this branch.  I can confirm that all fields are bit-for-bit except `grid_imask`, which no longer has missing values with this branch.

xylar:
@jonbob, could you check out this branch and make sure it fixes the issue you emailed me about?

@matthewhoffman, you were the original author of this SCRIP (but not of the bug I'm fixing).  I just want to make sure you're okay with these changes.  It's a pretty high priority, since we need to create working SCRIP files for v2.

@vanroekel  and @mark-petersen, I'm keeping you in the loop because there may be existing mapping files you're using that are affected by this problem.  Please flag anyone else you think needs to be aware.

conda_package/mpas_tools/scrip/from_mpas.py
```diff
@@ -39,7 +39,7 @@ def scrip_from_mpas(mpasFile, scripFile, useLandIceMask=False):
     lonCell = fin.variables['lonCell'][:]
     latVertex = fin.variables['latVertex'][:]
     lonVertex = fin.variables['lonVertex'][:]
-    verticesOnCell = fin.variables['verticesOnCell'][:]
+    verticesOnCell = fin.variables['verticesOnCell'][:] - 1
```
xylar, line 42: I think it's always better to subtract 1 at the start to get a valid python index array instead of having to remember to do this each time the array is used.

mark-petersen: I agree. That is a nice way to do it every time to avoid confusion. BTW, what is floor zero in Germany again? They are just way too logical. I personally count on my fingers, and I don't have a finger named zero :).

mark-petersen (approved):
@xylar thanks again for being proactive and your attention to detail. It looks like your testing is fine, since you have bfb results except for the fixed mask, and I am approving based on that. I appreciate being notified as a reviewer, just to see the traffic.

jonbob (approved):
Worked for me and fixed the issue we had been seeing with grid_imask. Thanks!

xylar:
@matthewhoffman, would you have time to review this today?  If not, that's okay.

matthewhoffman (approved):
@xylar , thanks for sorting these issues out.  Sorry for the slow turnaround on reviewing this.  It all looks good to me.

xylar:
@vanroekel has enough on his plate.  I'm taking him of as a reviewer.

@jonbob, @mark-petersen and @matthewhoffman, thanks very much for your reviews.

PR #354, opened 2020-10-02
Remove unneeded arguments from coastal_tools funcs

xylar:
* `origin` is not needed in `distance_to_coast()`
* `lon_grd` and `lat_grd` are identical to `lon` and `lat` in `compute_cell_width()`

### DIFF
--- a/conda_package/docs/ocean/coastal_tools.rst
+++ b/conda_package/docs/ocean/coastal_tools.rst
@@ -116,7 +116,6 @@ Next, the distance to the coastal contours is computed using
 :py:func:`mpas_tools.ocean.coastal_tools.distance_to_coast()` with the
 following values from ``params``:
 
-* ``'origin'`` - A lon and lat point---no longer used in the code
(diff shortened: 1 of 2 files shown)


xylar:
This is based on #349 and will need to be rebased for clarity once that gets merged.

xylar:
@sbrus89, if you could take a look at just the last commit in this PR and make sure it is okay, that would be great!  I grepped for both modified functions on https://github.com/MPAS-Dev/MPAS-Model/tree/ocean/develop and didn't get any hits so I think the only calls are the ones within https://github.com/MPAS-Dev/MPAS-Tools/pull/354/files#diff-0f20808562e2a380173337a8a5678c98R254 that I modified to remove the unneeded arguments.

xylar:
@sbrus89, this is rebased and ready to review.  The main thing would be to verify that removing these arguments will do no harm.

xylar:
@sbrus89, I'm going to go ahead a merge this.  We can handle any problems that emerge in COMPASS in future releases.

PR #356, opened 2020-10-08
Updates to MALI mesh interpolation script

matthewhoffman:
A number of upgrades to the MALI interpolation script:
* replace 2 deprecated python usage (optparse and time.clock)
* support multiple time levels (thanks @trhille )
* support variables with nVertInterfaces dimension
* add a number of new fields to dictionary of supported fields
* add ability to interpolate only variables included in a list on the command line


trhille (approved):
This looks great. I have no suggestions.

PR #358, opened 2020-10-17
Fix masking for signed distance functions

xylar:
Previously, we were buffering shapes by a full grid cell, but this was only necessary because we had the wrong origin for
`rasterio`'s `rasterize()` function.  If the origin is offset by half a pixel to account for the fact that `rasterio` pixel corners are half a pixel offset from the lon/lat that we want (at pixel centers), we only need to buffer shapes by a tiny amount (say `1e-10` degrees) to ensure that lon/lat points on the domain boundary are included within shapes that also touch the domain boundaries.


xylar:
I tested this on the SORRM and WC14 meshes, and found that the lon/lat plots of resolution were visually indistinguishable so I don't think this will change much.  But `nCells` is slightly different so this is definitely a non-bit-for-bit change, as expected.

PR #361, opened 2020-10-19
slight refactor to explose the core work as a method

sterlingbaldwin:
I tested this against output created by the previous version, and they appear to give identical results. The only change is to take the meat of main() and put it inside a function, so that it can be imported and called without needing to invoke the main and pass arguments via the command line. Also fixed a minor bug with args.inFileName vs args.inputFileName.


xylar:
@sterlingbaldwin, I don't think this is going to be useful because these are scripts that you can't import from the `mpas_tools` package.  I've been working on doing this the "right" way this evening and will have PR in a moment.

sterlingbaldwin:
ok, works for me.

PR #365, opened 2020-10-20
Add support for "unified" land-ocean meshes

dengwirda:
- Add `featureTagCell` + `featureTagVertex` to `jigsaw-to_netcdf.py`, and `mesh-conversion-tools` to enable labelling of cells/triangles via integer ID tags.
- New `tags` used to mark cells associated with geographic features: watersheds, rivers, basins, etc...
- Read jigsaw meshes via `jigsawpy` tools, to ensure consistent import of full `jigsaw_msh_t` data structure.
- Add `subtract_critical_passages` and `mask_reachable_ocean` to enable consistent culling of "unified" land-ocean meshes.
- See: [DOE-ICoM/icom-mesh/util/saveesm.py](https://github.com/DOE-ICoM/icom-mesh/tree/dev/util) for additional details.

### DIFF
--- a/conda_package/mpas_tools/ocean/coastline_alteration.py
+++ b/conda_package/mpas_tools/ocean/coastline_alteration.py
@@ -4,6 +4,12 @@ from __future__ import absolute_import, division, print_function, \
 import numpy
 import xarray
 
+from scipy.spatial import distance
+from scipy.sparse import csr_matrix
+from scipy.sparse.csgraph import connected_components
+
+from mpas_tools.mesh.creation.util import lonlat2xyz
+
 
 def add_critical_land_blockages(dsMask, dsBlockages):
     """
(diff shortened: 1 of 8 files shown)


dengwirda:
@xylar, hopefully this cleaned-up version is ready to go, but let me know if there are any other issues. Thanks for your help with this!

xylar:
@matthewhoffman, @mgduda, @mark-petersen, @akturner, these changes are needed for the ICoM project, which would like to have a unified ocean and land mesh, and would like to have a field as part of the mesh that can tag which cells and vertices are part of the ocean vs. the land.  Your reviews would be appreciated.

@dengwirda is proposing to increment the version of the MPAS mesh spec to 1.1 to include two new fields, `featureTagCell` and `featureTagVertex`.  This PR would update the mesh conversion tools to output those fields.  They would be all zeros if they are not provided in the input mesh.

xylar:
@dengwirda and @mgduda, if you don't know one another, it might be worth having an MPAS developers' telecon on which you could meet one another.  (That would be useful anyway.)  This PR could be one of the topics.

It couldn't be next week because of the ESMD meeting we're having but maybe the week after?

dengwirda:
To follow-up re: the new "tag" variables: as well as partitioning meshes into land + ocean pieces, we're also looking to tag cells / triangles associated with geometric features --- cells lying along rivers, triangles contained within watersheds, etc. 

This is to support new unstructured land + river models in the ICoM project, where extracting such structures are useful. This is also to enable integration with regional models (`ATS`, `FVCOM`, etc) which would run on sub-meshes embedded within the global grid.

The cell / vertex tags are integer ID's passed through from `jigsaw`, where such partitioned, boundary-aligned meshes can be built.

matthewhoffman:
@dengwirda , is there documentation of an example of how to use feature tags?

dengwirda:
@matthewhoffman, I don't have a test case or docs in MPAS-Tools itself yet, but [this case](https://github.com/DOE-ICoM/icom-mesh/tree/master/mesh/delaware_local_2d) from the icom-mesh workflow could be used to verify. This meshes a Delaware watershed, and tags river-aligned cells via `featureTagCell`. There are instructions in the [main readme](https://github.com/DOE-ICoM/icom-mesh) re: configuring icom-mesh, though it'd be necessary to build MPAS-Tools direct from this PR (rather than using the conda package).

land-river-mesh

conda_package/mpas_tools/ocean/coastline_alteration.py
```diff
@@ -35,6 +41,135 @@ def add_critical_land_blockages(dsMask, dsBlockages):
     return dsMask
 
 
+def subtract_critical_passages(dsMask, dsPassages):
+    """
+    Subtract the masks associated with one or more passages from the land
+    mask.
+
```
xylar, line 48: @dengwirda, is this different from adding a critical land blockage (see the function above)?

conda_package/mpas_tools/ocean/coastline_alteration.py
```diff
@@ -35,6 +41,135 @@ def add_critical_land_blockages(dsMask, dsBlockages):
...
+    return dsMask
+
+
+def mask_reachable_ocean(dsMesh, dsMask, fcSeed):
+    """
+    Return a new land mask that ensures all ocean cells are "reachable" from
+    (at least) one of the ocean points. Isolated patches of ocean cells will
+    be added to the land mask.
+
```
xylar, line 79: Just for clarification, is this just intended to be a simpler python version of the same capability that already exists in the C++ mask creator?  Or does this have different functionality?

xylar (changes requested):
@dengwirda, I'm sorry for not getting to this sooner.

I think this PR includes many different kinds of changes that need to be broken up into separate PRs.

I would include the changes to the C++ code in one PR, and I would separate white-space changes into a separate commit of that PR from the substantive code changes.  That PR will need to reviewed by everyone with a stake in the MPAS mesh specification, which will mean it needs a lot more scrutiny than the rest of the changes in the current PR.

Any changes under `conda_package` should be in another PR (or perhaps multiple others) that I can review without the others needing to be involved.  It seems like there are at least four conceptually distinct changes in this regard:
* clean-up of the recipe
* clean-up of how jigsaw output is converted to NetCDF
* changes to support the new tagging capability
* new functionality related to land masking
I would ideally have those in 4 distinct PRs but could imagine lumping the recipe clean-up in with one of the others since it's really minor.

xylar:
@dengwirda, what is he status of this PR at this point?  Should it be closed so we can aim for a different strategy or do you want to revisit this approach?

dengwirda:
@xylar, I think I've found a way to do this kind of re-indexing at the python level after the fact, so I'll close this now --- thanks.

xylar:
Great!

PR #368, opened 2020-10-30
Fix out-of-bounds issue with bilinear interp

xylar:
This merge fixes a bug in the bilinear interpolation routine in in `mpas_tools.mesh.interpolation`.  The specific case where an output point is exactly at the upper bound of the input grid needs special treatment and had an off-by-one index check.


xylar:
## Testing

I am using this function in a COMPASS test case (for a 1 km version of the ISOMIP+ test cases) where this error occurred.  Bilinear interpolation is correct with the bug fix but raises an out-of-bounds exception without it.

PR #370, opened 2020-11-01
Add _FillValue support to paraview extractor

xylar:
Up to now, it was looking only for ``missing_value`` and defaulting to a fill value that doesn't seem to work.


xylar:
## Testing

I was having problems with fill values not getting substituted in some COMPASS work I am doing.  After this fix, the substitution works as expected.

conda_package/mpas_tools/viz/paraview_extractor.py
```diff
@@ -1757,10 +1757,12 @@ def read_field_with_dims(field_var, dim_vals, temp_shape, outType,
         return field  # }}}
 
     field_var = get_var(var_name, mesh_file, time_series_file)
-    try:
+    if 'missing_value' in field_var.ncattrs():
         missing_val = field_var.missing_value
-    except AttributeError:
-        missing_val = -9999999790214767953607394487959552.000000
```
xylar, line 1763: I don't think we want to assume a fill value when no attribute is set.

PR #372, opened 2020-11-09
Remove compass metapackage

xylar:
It was moved to https://github.com/MPAS-Dev/compass, see https://github.com/MPAS-Dev/compass/pull/5


xylar:
From now on, it will make sense to maintain the metapackage within the new repo.

PR #376, opened 2020-11-15
Add depth dimension as arg to compute_zmid

xylar:
This will make the function useful for computing `zMid` in other contexts (e.g. COMPASS), not just CMIP6 data prep.

### DIFF
--- a/conda_package/mpas_tools/ocean/depth.py
+++ b/conda_package/mpas_tools/ocean/depth.py
@@ -14,16 +14,16 @@ def compute_depth(refBottomDepth):
 
     Parameters
     ----------
-    refBottomDepth : ``xarray.DataArray``
+    refBottomDepth : xarray.DataArray
         the depth of the bottom of each vertical layer in the initial state
         (perfect z-level coordinate)
 
     Returns
     -------
-    depth : ``xarray.DataArray``
+    depth : numpy.ndarray
         the vertical coordinate defining the middle of each layer
 
-    depth_bnds : ``xarray.DataArray``
+    depth_bnds : numpy.ndarray
         the vertical coordinate defining the top and bottom of each layer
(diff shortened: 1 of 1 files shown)


xylar:
@mark-petersen, with this change this function might be useful in COMPASS as a replacement for computing zMid in each test case from scratch.  I'm thinking specifically of: https://github.com/mark-petersen/MPAS-Model/blob/internal_tide_test_case/testing_and_setup/compass/ocean/internal_tide/3km/RK4/add_initial_state.py#L141-L147

conda_package/mpas_tools/ocean/depth.py
```diff
@@ -14,16 +14,16 @@ def compute_depth(refBottomDepth):
 
     Parameters
     ----------
-    refBottomDepth : ``xarray.DataArray``
+    refBottomDepth : xarray.DataArray
```
mark-petersen, line 17: What was the double backquote used for before, and why was it removed? I'm trying learn for myself.  Looks like a markdown thing the for documentation. I also just learned that a single backquote is an alias for `repl( )` in python, an unrelated fact.

mark-petersen (approved):
Looks good. If it works for you, please merge.

conda_package/mpas_tools/ocean/depth.py
xylar: There are in RestructuredText, which is a little different from markdown.

The double back-quotes are for literal strings, like single back-quotes in markdown.  But `sphinx` (the package that processes the documentation, including these docstrings) can link to `xarray` documentation if we don't use the back-quotes but not if we do.

I have never seen single back-quotes used in Python, nor have I ever used `repl()`.  Sounds fancy but potentially confusing...

PR #379, opened 2020-11-23
Switch MOC basins and transects to use aggregation from geometric_features

xylar:
The new approach takes advantage of the new ``moc()`` aggregation function from ``geometric_features``.

### DIFF
--- a/conda_package/mpas_tools/ocean/moc.py
+++ b/conda_package/mpas_tools/ocean/moc.py
@@ -5,10 +5,7 @@ import xarray
 import numpy
 import logging
 import sys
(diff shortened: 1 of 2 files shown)


xylar:
This PR requires a new `geometric_features` release after https://github.com/MPAS-Dev/geometric_features/pull/154 gets merged.

xylar:
## Testing

I ran the EC15to60E2r4 test case in COMPASS with a test build of this package.  COMPASS was able to produce the MOC region masks as usual with this code during the `files_for_e3sm` phase of the test case.

PR #381, opened 2020-12-14
Landice/temp outside ice mask exo mpas conversion

trhille:
This PR adds a few lines to the landice script conversion_exodus_init_to_mpasli_mesh.py that extend reasonable temperatures beyond the MPAS ice mask.


matthewhoffman:
@trhille , did you do a sanity check on the results from this?  If so, I'll improve by inspection.  If not, I'll actually bother to run it. :)

trhille:
Yes, I did.

PR #382, opened 2020-12-23
In write_netcdf, make Time an unlimited dimension

xylar:
Otherwise, MPAS-Model seems to have trouble reading these files.

### DIFF
--- a/conda_package/mpas_tools/io.py
+++ b/conda_package/mpas_tools/io.py
@@ -54,6 +54,11 @@ def write_netcdf(ds, fileName, fillValues=netCDF4.default_fillvals,
 
     update_history(ds)
 
+    if 'Time' in ds.dims:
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I have used this in my COMPASS 1.0 prtotype (https://github.com/MPAS-Dev/compass/pull/28) and with this change the files I write out with `write_netcdf` have `Time` as an unlimited dimension, whereas they were not before.  When these files are read into MPAS-Model the memory footprint is substantially reduced compared with the version without the unlimited dimension (which crashes my laptop after using the full memory).

I do not anticipate negative side effects of this change, since I think we would always want to produce MPAS files with `Time` as an unlimited dimension.  But more testing (e.g. in the current COMPASS) should be performed to be sure.

PR #385, opened 2021-01-01
Add support for logging (e.g. to log files)

xylar:
This merge adds a logging module to the conda package that handles making a logger that outputs to sdtout/stderr.  It also supports running a subprocess and redirecting the output to the logger.

Logging support has been added to the JIGSAW driver (along with a bit of clean-up).

The progress bar in the paraview extractor has been made optional.

Logging support has been added to both the ocean and general mesh builder.

### DIFF
--- a/conda_package/mpas_tools/logging.py
+++ b/conda_package/mpas_tools/logging.py
@@ -1,0 +1,187 @@
+import sys
+import logging
+import subprocess
+
+
+def check_call(args, logger):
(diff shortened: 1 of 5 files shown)


xylar:
## Testing

This has been tested in both the new compass 1.0 package and the current compass setup, and output is as expected (redirecting to logging files when requested, going to stdout/stderr otherwise).

PR #386, opened 2021-01-06
Add check for required time dimension if a range of times is included

matthewhoffman:
The interp script can optionally take a range of time levels to
interpolate.  Currently, if the expected time dimension is not in the
source file, the script will just interpolate the initial time level
from the source into all time levels of the destination.  This commit
fixes this behavior by aborting with an error in that situation.


trhille:
I have confirmed that this throws an error when there is no Time dimension, and works for multiple time levels when there is a Time dimension.

trhille (approved):

PR #396, opened 2021-03-02
Generalize angleEdge for planar meshes in MpasMeshConverter

matthewhoffman:
In MpasMeshConverter, the planar mesh calculation for angleEdge assumes
the connectivity of periodic_hex such that angleEdge would always be
between 0 and 180.

This commit eliminates that assumption and makes the angleEdge
calculation applicable to any mesh.  This is necessary for angleEdge to
be correct for planar meshes made with JIGSAW, for example.

### DIFF
--- a/mesh_tools/mesh_conversion_tools/mpas_mesh_converter.cpp
+++ b/mesh_tools/mesh_conversion_tools/mpas_mesh_converter.cpp
@@ -2159,6 +2159,7 @@ int buildAngleEdge(){/*{{{*/
 
 			normal = cell_loc2 - cell_loc1;
 			angleEdge.at(iEdge) = acos( x_axis.dot(normal) / (x_axis.magnitude() * normal.magnitude()));
+			if (cell_loc2.y < cell_loc1.y) angleEdge.at(iEdge) = 2.0 * M_PI - angleEdge.at(iEdge);
 		} else {
 
 			np = pntFromLatLon(edges.at(iEdge).getLat()+0.05, edges.at(iEdge).getLon());


mesh_tools/mesh_conversion_tools/mpas_mesh_converter.cpp
```diff
@@ -2159,6 +2159,7 @@ int buildAngleEdge(){/*{{{*/
 
 			normal = cell_loc2 - cell_loc1;
 			angleEdge.at(iEdge) = acos( x_axis.dot(normal) / (x_axis.magnitude() * normal.magnitude()));
+			if (cell_loc2.y < cell_loc1.y) angleEdge.at(iEdge) = 2.0 * M_PI - angleEdge.at(iEdge);
```
xylar, line 2162: This is definitely fine.  It would be more efficient to use `atan2` (http://www.cplusplus.com/reference/cmath/atan2/)

```
angleEdge.at(iEdge) = atan2(normal.y, normal.x);
```
Up to you.

trhille (approved):
Looks good to me!

matthewhoffman:
image
A before (left) and after (right) image for a variable resolution mesh generated by JIGSAW.  

- blue dots are cell centers
- red dots are vertices
- green X's are edge midpoints
- labels are the angleEdge values printed here in degrees (radians in code)

Note that angleEdge is defined as the CW angle between the eastward direction and the vector connecting the two cell centers straddling an edge.  (However the choice of which cell is 1 or 2 is arbitrary, so there are two possible angle values (differing by 180) - so this image does not make it possible to determine if the correct one is used.)

mesh_tools/mesh_conversion_tools/mpas_mesh_converter.cpp
matthewhoffman: Thanks, @xylar .  I'd rather not change anything at this point after the hours of debugging and careful testing that got to this point, but if you prefer to change it, we can.

xylar: No, that's fine.  The code for spherical meshes is also pretty inefficient.  If anyone cared to work on efficiency, there's plenty of low hanging fruit here...

matthewhoffman: Ok, thanks for pointing that out.  After the many hours of debugging MPAS model code only to find this bug here, I'm happy just moving forward.

xylar (approved):
I tested with a planar hex grid where periodicity had been removed.

I ran the culled mesh through the mesh converter, and I also rotated the mesh by 180 degrees and ran it through the mesh converter.  I did this using the old and the new code.

First, here are the results for `angleEdge` with the old code:
normal:
angle_edge_old
rotated by 180 degrees:
rotated_angle_edge_old
diff:
diff_old

Here is the same with the new code:
angle_edge_new
rotated_angle_edge_new
diff_new

With the new code, differences are always +/- `pi`.  Also, the angle of boundary edges are incorrect in the old code even without rotation -- any boundary edges that are below the valid cell center should have negative angles, which is only true for the new code.

mesh_tools/mesh_conversion_tools/mpas_mesh_converter.cpp
dengwirda: Yes, visually this looks good to me.
I don't think any changes are needed, but I do agree with @xylar that `atan2` is perhaps the best way to do this, which should give the correctly signed angle in `[-pi, +pi]`.

dengwirda (approved):

xylar:
@matthewhoffman, merge when you're ready.  I can make a new release of the conda package whenever you like.

matthewhoffman:
Thanks, Xylar.  We don't have any urgent need for this fix, so hopefully a conda package update can wait until there are other needs.

xylar:
Sounds good.  It's not a ton of work to release a package but it's a pain to update the various conda environment we use.  So I'll wait on all of that until another need arises.  Thanks for finding and fixing this bug!

PR #401, opened 2021-03-31
Update MOC southern boundary to not require region groups

xylar:
This will make it compatible with the python-based mask creator.

Fix some typos and formatting issues.


xylar:
## Testing

This was tested in MPAS-Analysis with QU240wLI and QU480.  Further testing will happen after the `mpas_tools` package gets updated.

PR #403, opened 2021-04-03
 Fix periodicity in python transect masks 

xylar:
Periodic images of polygons at the dateline were being updated in an incorrect order, causing transects that cross the dateline to have segments in arbitrary locations on the dateline.

This merge also includes some minor clean-up related to line length.


xylar:
## Testing

I made a testing transect through the Bering Strait and found that the mask is contiguous with this fix but not without it.

I have tested this along with changes in #404 in MPAS-Analysis on transport transects from simulations with the QU480, QU240wLI, SOwISC10to60E2r4 and WC14to60E2r3 meshes.

PR #404, opened 2021-04-03
Add edge sign to python transect masks

xylar:
This is optional, since we don't need the edge sign for critical passages and land blockages.

This capability needs a new package, `python-igraph`, for handling the graph of edges and vertices in a transect.

### DIFF
--- a/conda_package/docs/mesh_conversion.rst
+++ b/conda_package/docs/mesh_conversion.rst
@@ -350,10 +350,13 @@ The resulting variables are:
   - ``transectNames(nTransects, string64)`` - the names of the transects
 
 We don't currently provide cell, edge or vertex indices (e.g.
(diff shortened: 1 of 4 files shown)


xylar:
This PR builds on #403

xylar:
## Testing

So far, I have created the transport transects used in MPAS-Analysis.  The transects produced by this tool are much closer to straight lines than those produced by `MpasMaskCreator.x`.  Where they overlap, I have verified that the edge sign is always identical, giving me confidence that I have the computation correct.

I have made a fake "Arctic boundary" transect that crosses the date line and several continents to test that both of these situations are handled correctly.  In the process, I discovered the bug in #403.

xylar:
## Examples

Old:
old_afr_ant

New:
new_afr_ant

Old:
old_aus

New:
new_aus

Old:
old_carib

New:
new_carib

PR #406, opened 2021-04-05
Add python mask creation for projection grids

xylar:
This will be useful for creating region masks on polar grids, which are how some observational datasets are stored.


xylar:
## Testing

This worked great with the utility function added in https://github.com/MPAS-Dev/MPAS-Analysis/pull/777

PR #407, opened 2021-04-05
Add region group info to MOC output

xylar:
This is needed by the MOC analysis member in MPAS-Ocean but isn't provided by the python region mask creator


xylar:
This is needed to handle remaining issues with https://github.com/MPAS-Dev/MPAS-Analysis/pull/793

xylar:
## Testing

I ran the Anvil and Chrysalis MPAS-Analysis test suites using this branch and https://github.com/MPAS-Dev/MPAS-Analysis/pull/794, which makes use of these changes.

I also tested simulations with larger meshes (WCAtl12to45E2r4 and WC14to60E2r3)

PR #410, opened 2021-04-08
In grid_rotate/Makefile, set the Fortran compiler using 'nc-config --fc'

mgduda:
This merge sets the Fortran compiler using `nc-config --fc` in the grid_rotate
top-level Makefile.

Previously, the name of the Fortran compiler was hard-wired to 'gfortran'. Even
in cases where grid_rotate was being built with the GNU compilers, this could
fail with undefined references, for example, if the NetCDF library was build
with parallel I/O support (in which case, we need to link with the MPI wrapper).

Now, the value of the FC variable in the top-level Makefile is set to the output
of 'nc-config --fc'.

One potential issue worth noting is that, while the Fortran compiler can change
depending on the output of nc-config, the Fortran compiler flags (FFLAGS) are
still hard-wired in the top-level Makefile.

### DIFF
--- a/mesh_tools/grid_rotate/Makefile
+++ b/mesh_tools/grid_rotate/Makefile
@@ -1,4 +1,4 @@
-FC = gfortran
+FC = $(shell nc-config --fc)
 FFLAGS = -m64 -O2 -ffree-line-length-none -ffree-form -Wall
 FCINCLUDES = $(shell nc-config --fflags)
 FCLIBS = $(shell nc-config --flibs)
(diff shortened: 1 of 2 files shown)


mgduda:
@MiCurry Good point regarding the `README` -- I'll push an update soon to bring the documentation up to date. Regarding the `system` calls, I think we could consider changing those to `execute_command_line` calls as part of PR #411 .

MiCurry (approved):

PR #411, opened 2021-04-08
Replace non-standard calls to getarg and system in grid_rotate

mgduda:
This merge replaces non-standard `getarg` and `system` calls in the grid_rotate program
with calls to the F2003 `get_command_argument` and F2008 `execute_command_line`
subroutines, respectively.

### DIFF
--- a/mesh_tools/grid_rotate/grid_rotate.f90
+++ b/mesh_tools/grid_rotate/grid_rotate.f90
@@ -99,14 +99,14 @@ contains
 
       ! Copy original file to output file
       copyCmd = "cp " // trim(filename) // " " // trim(newFilename)
-      copyStat = system(copyCmd)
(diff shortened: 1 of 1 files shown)


MiCurry (approved):

mgduda:
@MiCurry As mentioned in PR #410, we could add another commit to this PR to replace the nonstandard `system` calls with `execute_command_line` calls as part of a broader objective of replacing non-standard calls of all sorts (`getarg`, `system`). What do you think?

MiCurry:
@mgduda I think using this PR to remove `system` would be great! Two birds with one stone!

mgduda:
@MiCurry I've just added another commit to address the `system` calls, and I've updated the PR description, too.

MiCurry (approved):
Looks great! Ship-it! :shipit:

PR #412, opened 2021-04-08
Update culling script to python3

mark-petersen:
Simple script update. I don't usually use this script, but I'm starting to test some sea ice verification tests that point to this one. Now runs with python3.

### DIFF
--- a/mesh_tools/periodic_hex/mark_periodic_boundaries_for_culling.py
+++ b/mesh_tools/periodic_hex/mark_periodic_boundaries_for_culling.py
@@ -20,7 +20,7 @@ import numpy as np
 from optparse import OptionParser
 
 
-print "== Gathering information.  (Invoke with --help for more details. All arguments are optional)"
+print("== Gathering information.  (Invoke with --help for more details. All arguments are optional)")
 parser = OptionParser()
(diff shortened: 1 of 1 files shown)


mesh_tools/periodic_hex/mark_periodic_boundaries_for_culling.py
```diff
@@ -20,7 +20,7 @@
 from optparse import OptionParser
```
xylar, line 20: Before all the other imports, it would be good to add:
```
from __future__ import print_function
```
This way, the script will still work with python 2 in case @akturner is still using it that way.

mark-petersen: Good point. Added.

xylar (approved):
This worked great for me.
```
planar_hex --nx 10 --ny 20 --dc 1000 -o mesh.nc
./mark_periodic_boundaries_for_culling.py -f mesh.nc
MpasCellCuller.x mesh.nc culled_mesh.nc
MpasMeshConverter.x culled_mesh.nc better_culled_mesh.nc
```
However, this tool is quite redundant with features already in `planar_hex`.  It would be nice to stop having to maintain a bunch of redundant tools.

xylar:
@akturner, could you look into modifying whichever tests you have that use this tool and see if you can just `planar_hex` instead?

PR #416, opened 2021-05-01
Add make_partition_files

xylar:
This is an entry point in the conda package or a stub script in `mesh_tools/make_partition_files` for computing a mesh partition file (`graph.info`) from an MPAS mesh.

### DIFF
--- a/conda_package/mpas_tools/parallel.py
+++ b/conda_package/mpas_tools/parallel.py
@@ -1,4 +1,6 @@
 import multiprocessing
+import argparse
+import xarray
 
 
 def create_pool(process_count=None, method='forkserver'):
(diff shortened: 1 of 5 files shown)


xylar:
This has been ported over from `compass`: https://github.com/MPAS-Dev/compass/blob/master/compass/model.py#L137-L195

xylar:
@jonbob, could you let me know if this is a script you use in the process of creating initial conditions for E3SM?  If so, where do you get the version you use from?

@matthewhoffman, in legacy COMPASS, this script was only used by the landice core as far as I can tell.  Would you ever foresee using it outside of `compass`?  If not, maybe it's fine to leave it there rather than moving it over here?

xylar:
Sorry, I seem to have confused `make_partition_files.py`, which is already in MPAS-Tools for `make_graph_file.py`, which was in COMPASS.  I don't think we need this PR.

PR #417, opened 2021-05-01
Move `make_partition_files` to conda package and update to python 3

xylar:
This merge updates the `make_partition_files` tool:
* it is moved into the conda package so it will be available in other workflows like `compass` and E3SM-Init.
* An entry point (command-line tool) named `make_partition_files` is added to the conda package
* the old scrip `mesh_tools/processor_decompositions/make_partition_files.py` is a stub that points to the new entry point
* a few changes are made to support python 3 (print function and integer division)

The `metis` package has been added as a dependency to the `mpas_tools` package


xylar:
Sorry about the false alarm with #416.  This is a PR to bring `make_partition_files.py` up to date.

@jonbob, I'll post the same questions I added on Confluence:

The old version of `make_partition_files.py` here would appear to me to be buggy (e.g. it imported `numpy` but used `np` in the code)  and only worked for python2.  So if you use this script, I am wondering whether you use the version from this repo or if you have your own copy or what the source is.

In general (whether old or new version), this script seems overly complicated for what we typically do.  I distinguishes between blocks and processors, for example.  So before this gets merged, I want to find out if this is the right point to consider a simpler tool for this purpose.  The COMPASS function [`make_graph_file()`](https://github.com/MPAS-Dev/compass/blob/master/compass/model.py#L137-L195) would seem to be a simpler version of the same process.

jonbob:
@xylar - I do use this script for generating graph files. I've never had any problem with it, other than it is occasionally slow. I think I have a version checked out from MPAS-Tools last June, and just point to a version of metis I had built.

xylar:
@jonbob, does your version have the bug that it imports `numpy` but refers to `np`?  If so, I'm wondering how it even works?

jonbob:
It does have those handy features. I'm guessing it only works because I load an e3sm-unified conda environment when I use it?

xylar:
@jonbob, I think I understand now.  It has some (very naughty) `from netCDF4 import *` and `from pylab import *`.  One of those must bring in `numpy` as `np`.

This script is a *haught mess* but if it works for now, fixing it doesn't have to be a high priority.

jonbob:
It does work, so I'm happy to leave it to you to prioritize the haught messes!

PR #418, opened 2021-05-01
Add a yaml file describing the `mpas_tools` dev environment

xylar:
This will let developers install `mpas_tools` into a conda environment in a way that points to the local branch.  This way, changes can easily be made to the `mpas_tools` package on the fly while testing the package in exactly the way a user or downstream package would.  This will be especially useful for testing changes in `mpas_tools` and packages that depend on these changes at the same time.

Creating a test environment is as simple as:
```
    cd conda_package
    conda env create -f ./dev_environment.yaml
    conda activate mpas_tools_dev
    python -m pip install -e .
```
No `conda build` or `conda create --use-local`.

The documentation has been updated with this new approach.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,6 @@ ENV/
 .ropeproject
 
 .DS_Store
+
+# pycharm
+.idea
--- a/conda_package/.gitignore
+++ b/conda_package/.gitignore
@@ -1,0 +1,5 @@
+# Files created by "python -m pip install -e ."
+/landice/
+/mesh_tools/
+/ocean/
+/visualization/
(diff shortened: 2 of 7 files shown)


xylar:
In the near future, when a `dev_environment.yaml` is available for MPAS-Analysis, geometric_features, compass, etc., the following should be possible for developing two (or more) packages simultaneously in an environment I've called `mpas_dev`:
```
conda create -y -n mpas_dev python=3.8
conda activate mpas_dev

cd ~/mpas_work/tools/MPAS-Tools/conda_package
conda env update -n mpas_dev -f ./dev_environment.yaml
python -m pip install -e .

cd ~/mpas_work/analysis/MPAS-Analysis
conda env update -n mpas_dev -f ./dev_environment.yaml
python -m pip install -e .
```

xylar:
@matthewhoffman, @mark-petersen, @sbrus89, @dengwirda, @trhille, just making you aware of this update.  Please ping anyone else who I might have missed.

PR #420, opened 2021-05-03
Switch from `rasterio` to `inpoly` for signed distance masks

xylar:
`inpoly` is likely much faster, is easier to use, and requires far fewer dependencies.  `inpoly` also is aware of detecting points exactly on the boundary of the polygon, eliminating the need to expand the polygon to capture points right on the boundary.

Both the `affine` and `rasterio` dependencies are removed as they are no longer needed.

`rasterio` is not yet compatible with the latest `proj`, which is causing trouble for updating `mpas_tools` to the latest `libnetcdf` (it's a long story...).  A new release with this update will eliminate that problem.

### DIFF
--- a/conda_package/mpas_tools/mesh/creation/signed_distance.py
+++ b/conda_package/mpas_tools/mesh/creation/signed_distance.py
@@ -95,14 +95,14 @@ def mask_from_geojson(fc, lon_grd, lat_grd):
 
     points = np.vstack([Lon.ravel(), Lat.ravel()]).T
 
-    in_shape, on_shape = inpoly2(points, nodes, edges)
+    in_shape, _ = inpoly2(points, nodes, edges)
 
-    mask = np.logical_or(in_shape, on_shape).reshape(Lon.shape)
+    mask = in_shape.reshape(Lon.shape)
     return mask
 
 
-def distance_from_geojson(fc, lon_grd, lat_grd, earth_radius, nn_search='flann',
-                          max_length=None):
+def distance_from_geojson(fc, lon_grd, lat_grd, earth_radius,
+                          nn_search='flann', max_length=None):
     # {{{
     """
     Get the distance for each point on a lon/lat grid from the closest point
@@ -182,8 +182,8 @@ def distance_from_geojson(fc, lon_grd, lat_grd, earth_radius, nn_search='flann',
         raise ValueError('Bad nn_search: expected kdtree or flann, got '
                          '{}'.format(nn_search))
 
-    # Convert  backgound grid coordinates to x,y,z and put in a nx_grd x 3 array
-    # for kd-tree query
+    # Convert background grid coordinates to x,y,z and put in a nx_grd x 3
+    # array for kd-tree query
     Lon_grd, Lat_grd = np.meshgrid(lon_grd, lat_grd)
     X_grd, Y_grd, Z_grd = lonlat2xyz(Lon_grd, Lat_grd, earth_radius)
     pts = np.vstack([X_grd.ravel(), Y_grd.ravel(), Z_grd.ravel()]).T
@@ -251,6 +251,7 @@ def _interpy(lat, x, y):
     lat_pixels = np.arange(nlat, dtype=float)
     y = np.interp(y, lat, lat_pixels)
     return x, y
+
 
 def _add_poly(poly, edges, nodes):
     node_count = len(nodes)
--- a/conda_package/mpas_tools/mesh/creation/signed_distance.py
+++ b/conda_package/mpas_tools/mesh/creation/signed_distance.py
@@ -1,21 +1,19 @@
-from __future__ import absolute_import, division, print_function, \
-    unicode_literals
-
 import numpy as np
 import pyflann
 from scipy import spatial
 import timeit
-from rasterio.features import rasterize
-from affine import Affine
 import shapely.geometry
 import shapely.ops
 from functools import partial
+from inpoly import inpoly2
+
 from geometric_features.plot import subdivide_geom
+
 from mpas_tools.mesh.creation.util import lonlat2xyz
 
 
 def signed_distance_from_geojson(fc, lon_grd, lat_grd, earth_radius,
-                                 max_length=None, epsilon=1e-10):
+                                 max_length=None):
     """
     Get the distance for each point on a lon/lat grid from the closest point
     on the boundary of the geojson regions.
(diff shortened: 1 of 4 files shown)


xylar:
## Testing

I made the base SOwISC12tto60 mesh with `rasterio` (old way) and `inpoly` (new way).

rasterio:
cellWidthGlobal

inpoly:
cellWidthGlobal

I didn't bother to compute timing because computing the mask is trivial compared to computing distance (both before and after this change).  But I did take a careful look at the mask in paraview and it looks exactly right.

xylar:
@dengwirda, all I'm hoping for in a review is a quick visual inspection that nothing looks obviously wrong or potentially problematic with my usage of `inpoly2` here.  If you want to run tests of your own, feel free.

conda_package/mpas_tools/mesh/creation/signed_distance.py
```diff
@@ -99,41 +76,33 @@ def mask_from_geojson(fc, lon_grd, lat_grd, max_length=None, shapes=None,
...
 
-    raster_shapes = []
-    for shape in shapes:
-        # expand a bit to make sure we hit the edges of the domain
-        shape = shape.buffer(epsilon)
-        raster_shapes.append((shapely.geometry.mapping(shape), 1.0))
+    in_shape, on_shape = inpoly2(points, nodes, edges)
 
-    mask = rasterize(raster_shapes, out_shape=(nlat, nlon), transform=transform)
+    mask = np.logical_or(in_shape, on_shape).reshape(Lon.shape)
```
dengwirda, line 100: I believe you should be able to get away without the `or` here, as `inpoly2` should return `in_shape` = `in_shape or on_shape` internally.

Additionally, just for reference, the `on_shape` output should be `TRUE` for points that lie within some tolerance of the boundaries, where "within" means that a manhattan-style dist. to boundary is small compared to the overall `xy`-length of the polygon. If it's necessary to detect boundary points exactly it may be worth discussing a little more.

conda_package/mpas_tools/mesh/creation/signed_distance.py
```diff
@@ -99,41 +76,33 @@ def mask_from_geojson(fc, lon_grd, lat_grd, max_length=None, shapes=None,
     print("Mask from geojson")
     print("-----------------")
 
-    if shapes is None:
-        shapes = _subdivide_shapes(fc, max_length)
```
dengwirda, line 103: Is it safe to remove this? I don't know the details as well as you @xylar, but does this impact the "spherical-ness" of the polygons, if they're specified as relatively coarse arcs in lon-lat space?

conda_package/mpas_tools/mesh/creation/signed_distance.py
xylar: Okay, thanks!  I'll clean that up.

conda_package/mpas_tools/mesh/creation/signed_distance.py
xylar: This would be useful if it were subdivided on the sphere, but it's just subdividing in lon/lat space (which is the space geojson files are defined in).  So I don't think it helps in any way to subdivide in that way.  It's important for the distance computation, though.

conda_package/mpas_tools/mesh/creation/signed_distance.py
xylar: Done.

conda_package/mpas_tools/mesh/creation/signed_distance.py
xylar: @dengwirda, it would be wonderful to replace the distance computation with better tools that I think you have, but that's beyond the scope of this PR, of course.

dengwirda: Yes okay, makes sense @xylar.
I agree that in a future (probably post-v2) PR we could revisit this, looking at what kind of arcs we may want in different places (i.e. [great vs small circles](https://en.wikipedia.org/wiki/Circle_of_a_sphere)) and whether a geodesic distance may be useful in certain applications (e.g. vs `jigsaw's` [`marche`](https://github.com/dengwirda/jigsaw-python/blob/b6e455c22696568f90a7428a49e03c980892c923/jigsawpy/libsaw.py#L776)).   

Will run a quick test and then approve this. All looks good I think!

dengwirda:
@xylar, I also tested with the full `region.geojson` file:
```
import time
import numpy
import xarray

from geometric_features import read_feature_collection

from mpas_tools.mesh.creation.signed_distance import mask_from_geojson
from mpas_tools.io import write_netcdf

if __name__ == '__main__':
   
    fc = read_feature_collection('region.geojson')

    dlon = 0.1
    dlat = 0.1
    nlon = int(360./dlon) + 1
    nlat = int(180./dlat) + 1

    lon = numpy.linspace(-180., 180., nlon)
    lat = numpy.linspace(-90., 90., nlat)   

    ttic = time.time()
    mask = mask_from_geojson(fc, lon, lat)
    ttoc = time.time()
    print("mask:", ttoc - ttic)

    ds = xarray.Dataset()
    ds['lon'] = ('lon', lon)
    ds['lat'] = ('lat', lat)
    ds['mask'] = (('lat', 'lon'), mask.astype(int))
    write_netcdf(ds, 'mask.nc')
```
which looks reasonable enough:
mask
although is a little different (at the boundaries) compared to the `rasterio` approach (new - old):
diff
It looks like the polygons are slightly "fatter" (i.e. more land) in the `rasterio` approach?

xylar:
@dengwirda, yes, we had to expand polygons slightly because they did not include the boundaries (i.e. ``in`` did not include ``on``), which was a problem.

xylar:
@dengwirda, if you're happy, please approve so I can merge and do a release tomorrow.

dengwirda:
@xylar, how important is it to maintain consistency across MPAS-Tools updates here? There are definitely some differences, this is a zoom of the amazon delta (`inpoly`, `rasterio`, `region.geojson`):
mask_zoom
It looks like the `inpoly` workflow is capturing the detail in the coastlines correctly, so I'm happy there, but if it's necessary to maintain closer consistency with the current MPAS-Tools I think we may need to look further?

xylar:
@dengwirda, I appreciate you being so thorough in your review.  I do not believe that masking is currently used in any workflow where the exact details of the shape are relevant.  For example, it is not currently used to handle the coastlines, that is another routine.  Instead, the current usage as far as I'm aware is for the sign part of signed distance functions and in a few cases for defining boundaries between basins, in which case the location of the boundary is intentionally placed over land so the exact location isn't important.

What is more, we do not try to guarantee that meshes are reproducible across versions of MPAS-Tools (or `mpas_tools`, the python package).  Indeed, if we also make updates to jigsaw and jigsaw-python soon, that won't be the case for sure.  So I think this is fine.

dengwirda (approved):

xylar:
Thanks, @dengwirda!

PR #422, opened 2021-05-07
Add a new utility to create ESMF and SCRIP files for MPAS-A

MiCurry:
This merge adds a Fortran utility that can generate ESMF and SCRIP files from an MPAS mesh. Although there already exists a utility that can create SCRIP files from an MPAS mesh, that utility uses Conda, which is not part of the MPAS-Atmosphere's current workflow.

Currently, this tool is used to generating SCRIP files for MPAS in CESM.

### DIFF
--- a/mesh_tools/mpas2esmf/.gitignore
+++ b/mesh_tools/mpas2esmf/.gitignore
@@ -1,0 +1,1 @@
+mpas2esmf
--- a/mesh_tools/mpas2esmf/Makefile
+++ b/mesh_tools/mpas2esmf/Makefile
@@ -1,0 +1,9 @@
+FC = $(shell nc-config --fc)
+FCINCLUDES = $(shell nc-config --fflags)
+FCLIBS = $(shell nc-config --flibs)
+
+all: mpas2esmf.f90
+	$(FC) -o mpas2esmf mpas2esmf.f90 ${FCINCLUDES} ${FCLIBS}
+
+clean:
+	rm -f mpas2esmf read_mesh.mod write_desc.mod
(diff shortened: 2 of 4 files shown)


MiCurry:
@xylar Althought I don't think there is, let us know if you think there will be a conflict with this tool and the `create_SCRIP_files` utility.

xylar:
@MiCurry, no, there is no relation between the two tools so there should also be no interference.  Thanks for asking!

MiCurry:
> @MiCurry Thanks for putting together a Makefile and PR for this. I have just a few minor suggestions:
> 
>     1. Should we add a `.gitignore` file with `mpas2esmf` as its only line?
> 
>     2. Should the `Makefile` set `FC`, e.g., based on the output of `$(shell nc-config --fc)` as we do for [grid_rotate](https://github.com/MPAS-Dev/MPAS-Tools/blob/master/mesh_tools/grid_rotate/Makefile), or is the expectation that the user will set `FC` on the command-line?

Those both sounds like good ideas! I wasn't aware of `nc-config --fc`. I'll add them both. Do you think its worthwhile to also have the same `FFLAGS` used as `grid_rotate`?

mgduda:
If we can avoid having to add `FFLAGS`, I think that's preferable, as it makes the default `Makefile` compatible with multiple Fortran compilers. With the `grid_rotate` tool, the `FFLAGS` only work with the GNU compiler, and users need to manually change these when switching to other Fortran compilers.

MiCurry:
@mgduda sounds good! I just pushed (force pushed) an update to this branch with the changes you requested. I also added a short README. Let me know if there is anything else.

mgduda (approved):

PR #424, opened 2021-05-26
Update `make_partition_file.py` to python 2 and 3

xylar:
This will likely be require as the default python moves to python 3 on most supported systems.

Print statements now have parentheses.  Integer division is now with `//`.

### DIFF
--- a/mesh_tools/processor_decompositions/README
+++ b/mesh_tools/processor_decompositions/README
@@ -16,13 +16,13 @@ Options:
                         Field to weight block partition file on.
 
 About:
-	This script is used to create weighted and unweighted hirearchical 
+	This script is used to create weighted and unweighted hierarchical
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I tested this in Linux with the `ECwISC30to60E2r1` with 8 procs and 8 blocks using `gpmetis` from conda-forge.  I tried both without weights and using `maxLevelCell` for weights.  I ran the old version of the script (from a python 2.7 environment) and verified with `diff` that all 5 files produced with the new and old versions of the script are identical.

I ran the new version of the script with both python 2.7 and 3.9.

xylar:
@jonbob, I'm hoping this will be an easier branch to review.  Could you let me know when you will have a chance to test this so I can merge?  Thanks in advance!

jonbob (approved):
worked great for me on anvil

xylar:
Thanks so much, @jonbob!

PR #426, opened 2021-06-08
Land Ice: modify script to create MPAS grid

mperego:
Modified file create_grid_landice_grid_from_generic_MPAS_grid.py to
* always create muFriction, initializated at 1e8 [yr/m]
* change the default of effectivePressure to 1 [Pa] instead of 1e8 [Pa]
* always create beta
* deprecate the --beta and --effecPress command line options

In the future, beta will be removed from MALI.  After that transition is complete, this script will be modified to remove beta and the deprecated options.  The current version supports both backwards and forwards compatibility.


matthewhoffman:
@mperego , @trhille , do you think we should just add a `--ho` option that includes all of beta, muFriction, and effecPressure?  Or alternatively, have muFriction automatically included with the `--effecpress` option?

Or, maybe better yet, we make all of those 3 variable always included in a new mesh.  At this point, MALI is mature enough that most of the time we want those fields.  We never use the SIA setup for anything but testing.  The only time we might not want these fields is when running the subglacial hydro model on its own.  But in that case we generally are starting from the result of some optimization.

trhille:
@matthewhoffman , I agree that they should be included automatically. I think it's cleaner to remove them if needed than to specify that we want them 95% of the time.

matthewhoffman:
@mperego , I pushed a new commit that does what we discussed - makes beta, effectivePressure, and muFriction always added to new meshes, and also deprecates the --beta and --effecPress options.  Once we've cleanly made the transition in the whole workflow, we'll remove beta and those deprecated options.  (Or maybe we forget and never do it.)

Let me know if you are ok with this and I'll merge it.

cc: @trhille

mperego:
> @mperego , I pushed a new commit that does what we discussed - makes beta, effectivePressure, and muFriction always added to new meshes, and also deprecates the --beta and --effecPress options. Once we've cleanly made the transition in the whole workflow, we'll remove beta and those deprecated options. (Or maybe we forget and never do it.)
> 
> Let me know if you are ok with this and I'll merge it.
> 
> cc: @trhille

Looks good to me. Thanks!

trhille (approved):
This looks good and ready to merge to me.

PR #430, opened 2021-07-11
Fix re-registering of colormaps

xylar:
Matplotlib will complain if we try to re-register a colormap that has already been registered.


xylar:
## Testing

I tested this in a script that generates several ocean meshes on the sphere.  With `master`, I got warnings about reregistering existing colormaps that are gone when I retested with this branch.

PR #431, opened 2021-07-13
Fix uninitialized values in kiteAreasOnVertex

xylar:
Previously, the MPAS mesh converter was not initializing all values in the temporary buffer for writing out `kiteAreasOnVertex`, meaning some areas for nonexistent cells had garbage values.  This can cause MPAS model to fail in debug mode with arithmetic errors.

### DIFF
--- a/mesh_tools/mesh_conversion_tools/mpas_mesh_converter.cpp
+++ b/mesh_tools/mesh_conversion_tools/mpas_mesh_converter.cpp
@@ -3067,6 +3067,9 @@ int outputVertexParameters( const string outputFilename) {/*{{{*/
 	// Build and write kiteAreasOnVertex
 	// TODO: Fix kite area for quads?
 	tmp_arr = new double[nVertices*vertexDegree];
+	for(i = 0; i < nVertices*vertexDegree; i++){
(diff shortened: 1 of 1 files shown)


xylar:
@mgduda, @akturner and @matthewhoffman. This is an issue that is messing up compass when we run in debug mode.  I'd like you all to review (either just by inspection or running some tests if you feel the need) since the mesh converter is one of the most important bits of shared infrastructure that we share.

matthewhoffman (approved):
Approved by inspection.

akturner:
@xylar: Looks good.

xylar:
Thanks @matthewhoffman and @akturner for reviewing so quickly.

mgduda (approved):
Approved by inspection. The use of an `int` loop index variable rather than a `size_t` seems fine, since `nVertices * vertexDegree` should be well under 2^31 - 1 for any meshes that we will generate for the foreseeable future.

xylar:
@mgduda, I'm guessing we'd have problems elsewhere too if meshes got that big. But worth having in mind. Thanks for the quick review!

PR #433, opened 2021-07-19
Update performance script to python 3, and minor corrections

mark-petersen:
Minor script updates. With these changes, the script produces performance scaling plots correctly.

### DIFF
--- a/ocean/performance_testing/performance_test.py
+++ b/ocean/performance_testing/performance_test.py
@@ -111,8 +111,8 @@ while i >= nprocs_min:
     sum = 0
 
     # Generate the log and graph files
-    subprocess.check_call(['./metis', 'graph.info', str(i)])
-    print "metis" + str(i) + "completed"
+    subprocess.check_call(['gpmetis', 'graph.info', str(i)])
+    print("gpmetis" + str(i) + "completed")
 
     for sample in range(nsamples_per_procnum):
         subprocess.check_call(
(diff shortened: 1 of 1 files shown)


ocean/performance_testing/performance_test.py
```diff
@@ -165,6 +165,6 @@
     figurenamepath = "figures_performance/" + resolution_name + '_' + \
         machine_long_name + '_' + str(nprocs_max) + "_" + timenow + ".png"
     plt.savefig(figurenamepath)
-    i = i / 2
+    i = int( i / 2 )
```
xylar, line 168: The preferred way to do this is:
```suggestion
    i = i // 2
```

mark-petersen:
In a run directory, the command
```
./performance_test.py 1024 32 grizzly 36 EC60to30
```
produces this plot:
EC60to30_grizzly_1024_20210719_081846

xylar:
@mark-petersen, if you want to make the one suggested change above, I'll approve by inspection. If you would like me to test, please let me know what typical usage looks like.

xylar:
@mark-petersen, based on your typical usage above (our messages crossed!), would you be willing to add `argparse` to this script so the arguments are named, rather than ordered and so there's help when you do `./performance_test.py  --help`.

Should we add this to the conda package?

xylar:
If the answer is no to all of the above, that's fine.

mark-petersen:
@xylar I agree about argparse. I had thought of that too. If you approve I'll merge this in now so students can use it, and argparse is a good exercise for a student to update.

mark-petersen:
@xylar no need to test, visual inspection is fine.

xylar (approved):
Looks good.

PR #434, opened 2021-07-28
Add argument to `write_netcdf` for the NetCDF engine

xylar:
Performance for the default `netcdf4` engine has found to be poor for large files on Chrysalis and Anvil, and `scipy` seems to perform better.  For now, `netcdf4` remains the default but the option to change is now provided.

To aid in this process, default values have been defined in the `mpas_tools.io` module that can be modified once, rather than
each time `write_netcdf()` is called.

The `planar_hex` tool and mask creation tools have been modified to take the format and engine as arguments, so that calling code can have more control of these if needed.

### DIFF
--- a/conda_package/mpas_tools/io.py
+++ b/conda_package/mpas_tools/io.py
@@ -7,8 +7,14 @@ from datetime import datetime
 import sys
 
 
-def write_netcdf(ds, fileName, fillValues=netCDF4.default_fillvals,
-                 format='NETCDF3_64BIT', char_dim_name='StrLen'):
+default_format = 'NETCDF3_64BIT'
+default_engine = None
+default_char_dim_name = 'StrLen'
+default_fills = netCDF4.default_fillvals
+
+
+def write_netcdf(ds, fileName, fillValues=None, format=None, engine=None,
+                 char_dim_name=None):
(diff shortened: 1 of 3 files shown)


xylar:
Python 3.6 testing is failing because `xarray` has dropped support.  We'll do the same in #435

xylar:
## Testing

As reported here https://github.com/MPAS-Dev/compass/pull/187#issuecomment-888613015, I ran the nightly ocean test suite successfully with this branch and https://github.com/MPAS-Dev/compass/pull/187 on Ubuntu with gnu and Chrysalis with intel, comparing against a baseline of the current master.

The `ssh_adjustment` step with the SOwISC12to60 mesh now works like a charm, so this does seem to be a good solution to https://github.com/MPAS-Dev/compass/issues/185.

xylar:
@mark-petersen and @matthewhoffman, I'm going to go ahead and merge this and make an MPAS-Tools release.  You can then test these changes as part of reviewing https://github.com/MPAS-Dev/compass/pull/187.  It will be tricky to test both this PR and that one at the same time so I'd prefer to do it that way, since I'm pretty confident these changes are working as expected.

PR #437, opened 2021-07-29
Rework file I/O in mesh converters for thread-safety

dengwirda:
This PR reworks the way the C++ `MpasMeshConverter.x` and `MpasCellCuller.x`, as well as the MPAS-Tools `conversion.py` wrapper, handles file I/O to eliminate the need to change into local directories to account for hard-coded relative file names (e.g. `graph.info`).

Eliminating the `os.getcwd()` + `os.chdir()` pattern in `conversion.py` makes these routines thread-safe, allowing them to be called across parallel threads.

These changes write all output files generated by the mesh-converter/culler (e.g. `graph.info`, `cellMapForward.txt`, etc) into the directory defined by the output filename.

### DIFF
--- a/mesh_tools/mesh_conversion_tools/mpas_cell_culler.cpp
+++ b/mesh_tools/mesh_conversion_tools/mpas_cell_culler.cpp
@@ -12,6 +12,8 @@
 
 #include "netcdf_utils.h"
 
+#include "string_utils.h"
+
 #define ID_LEN 10
 
 using namespace std;
@@ -53,10 +55,11 @@ int markEdges();
 int outputGridDimensions(const string outputFilename);
 int outputGridAttributes(const string inputFilename, const string outputFilename);
 int mapAndOutputGridCoordinates(const string inputFilename, const string outputFilename);
-int mapAndOutputCellFields(const string inputFilename, const string outputFilename);
+int mapAndOutputCellFields(const string inputFilename, const string outputPath, 
+                           const string outputFilename);
 int mapAndOutputEdgeFields(const string inputFilename, const string outputFilename);
 int mapAndOutputVertexFields(const string inputFilename, const string outputFilename);
-int outputCellMap();
+int outputCellMap(const string outputPath);
 /*}}}*/
 
 void print_usage(){/*{{{*/
@@ -95,6 +98,9 @@ int main ( int argc, char *argv[] ) {
 	int error;
 	string out_name = "culled_mesh.nc";
 	string in_name = "mesh.nc";
+    string out_path = "";
+    string out_file = "";
+    string out_fext = "";
 	vector<string> mask_names;
 	vector<int> mask_ops;
 
@@ -183,6 +189,8 @@ int main ( int argc, char *argv[] ) {
 		return 1;
 	}
 
+    file_part(out_name, out_path, out_file, out_fext);
+
 	srand(time(NULL));
 
 	cout << "Reading input grid." << endl;
@@ -228,7 +236,7 @@ int main ( int argc, char *argv[] ) {
 	}
 
 	cout << "Mapping and writing cell fields and culled_graph.info" << endl;
-	if(error = mapAndOutputCellFields(in_name, out_name)){
+	if(error = mapAndOutputCellFields(in_name, out_path, out_name)){
 		cout << "Error - " << error << endl;
 		exit(error);
 	}
@@ -247,7 +255,7 @@ int main ( int argc, char *argv[] ) {
 
 	cout << "Outputting cell map" << endl;
 	if (outputMap) {
-		if(error = outputCellMap()){
+		if(error = outputCellMap(out_path)){
 			cout << "Error - " << error << endl;
 			exit(error);
 		}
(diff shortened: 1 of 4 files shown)


dengwirda:
To run additional tests (attached):
```
cd conda_package
conda env create -y -n mpas_tools_dev --file dev-spec.txt
conda activate mpas_tools_dev
python -m pip install -e .
cd ..
cd mesh_tools/mesh_conversion_tools
mkdir build
cd build
cmake -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DCMAKE_BUILD_TYPE=Debug ..
make
make install
```
```
python test_sphere.py
python test_planar.py
```
[test_mesh_converter.zip](https://github.com/MPAS-Dev/MPAS-Tools/files/6903180/test_mesh_converter.zip)

mesh_tools/mesh_conversion_tools/mpas_cell_culler.cpp
```diff
@@ -95,6 +98,9 @@ int main ( int argc, char *argv[] ) {
 	int error;
 	string out_name = "culled_mesh.nc";
 	string in_name = "mesh.nc";
+    string out_path = "";
+    string out_file = "";
+    string out_fext = "";
```
xylar, line 103:
```suggestion
	string out_path = "";
	string out_file = "";
	string out_fext = "";
```
These should be tabs, not spaces, to be consistent with existing file formatting (ugly as it may be)

mesh_tools/mesh_conversion_tools/mpas_cell_culler.cpp
```diff
@@ -183,6 +189,8 @@ int main ( int argc, char *argv[] ) {
 		return 1;
 	}
 
+    file_part(out_name, out_path, out_file, out_fext);
```
xylar, line 192:
```suggestion
	file_part(out_name, out_path, out_file, out_fext);
```
Again, tabs...

mesh_tools/mesh_conversion_tools/mpas_cell_culler.cpp
xylar: The same formatting issue is also in `mpas_mesh_converter.cpp`.

xylar (changes requested):
For now, just requesting a few formatting fixes.  I'll test this in compass tomorrow.  Thanks for the nice example tests and instructions!

mesh_tools/mesh_conversion_tools/mpas_cell_culler.cpp
dengwirda: Ah, right... no worries!

xylar:
@mgduda, @akturner, @matthewhoffman and @mark-petersen, this work is needed as part of working on supporting parallel tasks via Parsl.  If we make meshes in parallel with threads, it turns out that we can't be doing `chdir` type operations because these aren't thread-safe in python.

xylar:
@dengwirda, could you add to the PR description that the approach for modifying the C++ code is to use the directory of the output file also as the directory for the other files (e.g. graph.info)?  I think that's missing.

matthewhoffman (approved):
I'm just going to approve by inspection.  This makes sense and the changes look appropriate.  Thank for working on this Darren.

xylar:
@dengwirda, I tested this in `compass` on both my Linux laptop and Chrysalis.  I needed to build the conda package (so that conda-forge compilers were used instead of system compilers) rather than following the instructions above, which used system compilers.  When I did that, I got bit-for-bit identical results with the ocean nightly test suite for all tests.

akturner (approved):
Built and ran successfully for me

mark-petersen (approved):
I compiled the c++ MpasCellCuller and MpasMeshConverter on grizzly with both gnu and intel and tested them individually with a QU240 mesh. Everything appears to work fine. Thanks!

mesh_tools/mesh_conversion_tools/mpas_mesh_converter.cpp
```diff
@@ -166,6 +171,7 @@ int main ( int argc, char *argv[] ) {
 		return 1;
 	}
 
+    file_part(out_name, out_path, out_file, out_fext);
```
mgduda, line 174: If we would like to match existing formatting, the spaces here should be changed to a tab.

mgduda (changes requested):
I'll second @xylar's suggestion to fix the whitespace in `mpas_cell_culler.cpp` and `mpas_mesh_converter.cpp`.

mesh_tools/mesh_conversion_tools/mpas_mesh_converter.cpp
```diff
@@ -119,6 +121,9 @@ int main ( int argc, char *argv[] ) {
 	int error;
 	string out_name = "mesh.nc";
 	string in_name = "grid.nc";
+    string out_path = "";
+    string out_file = "";
+    string out_fext = "";
```
xylar, line 126:
```suggestion
	string out_path = "";
	string out_file = "";
	string out_fext = "";
```

mesh_tools/mesh_conversion_tools/mpas_mesh_converter.cpp
xylar:
```suggestion
	file_part(out_name, out_path, out_file, out_fext);
```

xylar:
@mgduda, I went ahead and fixed the white space since it seems like @dengwirda hasn't had time.  Could you give it another look to make sure you're okay with the changes otherwise?

xylar (approved):
Approving now that white space issues are fixed.

xylar:
@mgduda, is this something you might have time to try out soon?  If so, the timing would be good with other fixes we want to make for another MPAS-Tools release.

mgduda (approved):
Sorry about the delay!

xylar:
Thanks @mgduda!

PR #438, opened 2021-08-20
Fix error in MOC southern transect

xylar:
We need to make sure cells on the dummy edge at the end of the edge array are masked as out of range.

Addresses https://github.com/MPAS-Dev/compass/issues/218


xylar:
## Testing

I ran the `make_diagnostics_files` test case from `compass` on Chrysalis using this fix.  Before the fix, I saw the issue reported in https://github.com/MPAS-Dev/compass/issues/218 but with the fix I was able to create the MOC mask files for the EC30to60 mesh in `compass`.

PR #440, opened 2021-09-14
Landice/add mu friction to exo mpas conversion

trhille:
This adds the basal friction coefficient muFriction field to conversion_exodus_init_to_mpasli_mesh.py. muFriction is treated similarly to beta, and extrapolated using nearest-neighbors to fill the MPAS mesh outside of the Albany mesh.


matthewhoffman (commented):
Reviewing by inspection and based on Trevor's testing.

PR #442, opened 2021-09-22
 Switch from `pyflann` to scipy `KDTree`

### LINKED ISSUE #441: drop (`py`)`flann` from conda package
It seems that the `flann` package is not being maintained.  The last commits were in 2019: https://github.com/flann-lib/flann/commits/master
Web site links are broken.

This lack of support is likely to manifest itself in an inability to keep `mpas_tools` updated with the latest conda-forge libraries, which will negatively affect E3SM-Unified as well.  The current problem is with: https://github.com/conda-forge/flann-feedstock/pull/31

I suggest that we figure out how to move to a supported package instead.

xylar:
Anywhere there was the option to use both, there is now only one option (always 'kdtree').

This merge also updates ci builds of the conda package to use `libnetcdf 4.8.1`.

closes #441

### DIFF
--- a/conda_package/mpas_tools/mesh/creation/signed_distance.py
+++ b/conda_package/mpas_tools/mesh/creation/signed_distance.py
@@ -12,7 +12,7 @@ from mpas_tools.mesh.creation.util import lonlat2xyz
 
 
 def signed_distance_from_geojson(fc, lon_grd, lat_grd, earth_radius,
-                                 max_length=None):
+                                 max_length=None, workers=-1):
     """
     Get the distance for each point on a lon/lat grid from the closest point
     on the boundary of the geojson regions.
@@ -35,6 +35,10 @@ def signed_distance_from_geojson(fc, lon_grd, lat_grd, earth_radius,
         The maximum distance (in degrees) between points on the boundary of the
         geojson region.  If the boundary is too coarse, it will be subdivided.
 
+    workers : int, optional
+        The number of threads used for finding nearest neighbors.  The default
+        is all available threads (``workers=-1``)
+
     Returns
     -------
     signed_distance : numpy.ndarray
@@ -42,7 +46,8 @@ def signed_distance_from_geojson(fc, lon_grd, lat_grd, earth_radius,
        to the shape boundary
     """
     distance = distance_from_geojson(fc, lon_grd, lat_grd, earth_radius,
-                                     nn_search='flann', max_length=max_length)
+                                     nn_search='kdtree', max_length=max_length,
+                                     workers=workers)
 
     mask = mask_from_geojson(fc, lon_grd, lat_grd)
 
@@ -101,7 +106,8 @@ def mask_from_geojson(fc, lon_grd, lat_grd):
 
 
 def distance_from_geojson(fc, lon_grd, lat_grd, earth_radius,
-                          nn_search='kdtree', max_length=None):
+                          nn_search='kdtree', max_length=None,
+                          workers=-1):
     # {{{
     """
     Get the distance for each point on a lon/lat grid from the closest point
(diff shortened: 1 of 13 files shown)


xylar:
## Testing

I ran the nightly compass ocean test suite using a test build of this package on my Ubuntu laptop with gnu compilers.  It was bit-for-bit for all tests that ran (tests with more than 4 cores fail).  This isn't a terribly rigorous test, so I will do a similar test on Chrysalis later today or tomorrow.

xylar:
@sbrus89, do you have a plausible way to run any tests with this branch?  Would you need me to make a test build of the package and/or a development environment to test with?

@dengwirda, let me know if you would like to test this.  If not, that's fine.  I would still appreciate having you take a quick look at the code, particularly to make sure it doesn't make it harder for you to add support for geodesic distance in the future.

conda_package/mpas_tools/mesh/creation/signed_distance.py
```diff
@@ -191,12 +181,7 @@ def distance_from_geojson(fc, lon_grd, lat_grd, earth_radius,
     # Find distances of background grid coordinates to the coast
     print("   Finding distance")
     start = timeit.default_timer()
-    distance = None
-    if nn_search == "kdtree":
-        distance, _ = tree.query(pts)
-    elif nn_search == "flann":
-        _, distance = flann.nn_index(pts, checks=2000, random_seed=0)
-        distance = np.sqrt(distance)
+    distance, _ = tree.query(pts)
```
dengwirda, line 184: I believe it's possible to do [`tree.query(pts, workers=-1)`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.KDTree.query.html#scipy.spatial.KDTree.query) to have this run on whatever `threads=cores` are available, which may speed things up for big cases.

xylar: Yes, I think that's a good idea.  Let's add a parameter to each of these functions so there's user control if needed but make `-1` the default.

dengwirda: Yes, this sounds like a good plan!

xylar:
@dengwirda, I just added support for `workers` with the default being `workers=-1`.  Let me know how it looks.  I'll test tomorrow but I'm off to bed for now.

sbrus89:
> @sbrus89, do you have a plausible way to run any tests with this branch? Would you need me to make a test build of the package and/or a development environment to test with?

There are tests in legacy compass that use the `costal_tools.py` script. If you could create a developer environment, that would be great. Then I can run those tests.

xylar:
> If you could create a developer environment, that would be great. Then I can run those tests.

@sbrus89, sure.  What machine would you like to use?  Chrysalis?

sbrus89:
@xylar, Compy would be great, it's all the same to you. Otherwise Chrysalis is good too.

xylar:
Compy it is!  Not my preferred machine ;-)  But it's no harder for me to set things up there than anywhere else.

xylar:
@sbrus89, Compy is very slow but soon you should be able to run:
```
source /share/apps/E3SM/conda_envs/base/etc/profile.d/conda.sh
conda activate test_compass_0.3.0
```
And you should have the test environment you need.  As far as compiling MPAS-Ocean for legacy, you can try the following, which is based on the load script for the new compass:
```
source /share/apps/E3SM/conda_envs/base/etc/profile.d/conda.sh
conda activate test_compass_0.3.0

module purge
module load cmake/3.19.6
module load gcc/8.1.0
module load intel/19.0.5
module load intelmpi/2019u4
module load netcdf/4.6.3
module load pnetcdf/1.9.0
module load mkl/2019u5

export NETCDF=$(dirname $(dirname $(which nc-config)))
export NETCDFF=$(dirname $(dirname $(which nf-config)))
export PNETCDF=$(dirname $(dirname $(which pnetcdf-config)))

export PIO=/share/apps/E3SM/conda_envs/compass/system/compass_1.0.0/intel/impi/scorpio_1.2.1
export USE_PIO2=true
export HDF5_USE_FILE_LOCKING=FALSE
```

sbrus89 (approved):
Approved based on testing of `-o ocean -c hurricane -r USDEQU120at30cr10rr2 -t build_mesh` from legacy compass. The mesh builds as expected after the `pyflann` -> `KDTree` modifications to the `coastal_tools.py` script.  Performance did not seem to be affected at all.

culled_mesh

xylar:
Thanks @dengwirda and @sbrus89!

PR #444, opened 2021-10-04
Update CISM script to be compatible with python 3

hollyhan:
This merge
- fixes proj syntax.
- takes options for data rank (1 or 2).


mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
```diff
@@ -17,26 +22,37 @@
...
 # NOTE!!!!!!  egm08_25.gtx can be downloaded from:  http://download.osgeo.org/proj/vdatum/egm08_25/egm08_25.gtx  and the path in the projection specification line should point to it!
 #projections['gis-bamber'] = pyproj.Proj('+proj=stere +lat_ts=71.0 +lat_0=90 +lon_0=321.0 +k_0=1.0 +x_0=800000.0 +y_0=3400000.0 +geoidgrids=./egm08_25.gtx')
-projections['gis-bamber'] = pyproj.Proj('+proj=stere +lat_ts=71.0 +lat_0=90 +lon_0=321.0 +k_0=1.0 +x_0=800000.0 +y_0=3400000.0 +ellps=WGS84')  # This version ignores the vertical datum shift, which should be a very small error for horizontal-only positions
+#projections['gis-bamber'] = pyproj.Proj('+proj=stere +lat_ts=71.0 +lat_0=90 +lon_0=321.0 +k_0=1.0 +x_0=800000.0 +y_0=3400000.0 +ellps=WGS84')  # This version ignores the vertical datum shift, which should be a very small error for horizontal-only positions
+
+# GIMP projection: This is also polar stereographic but with different standard parallel and using the WGS84 ellipsoid.
+#projections['gis-gimp'] = pyproj.Proj('+proj=stere +lat_ts=70.0 +lat_0=90 +lon_0=315.0 +k_0=1.0 +x_0=0.0 +y_0=0.0 +ellps=WGS84')
+
+
```
hollyhan, line 30: fix white space

hollyhan:
```suggestion
```

hollyhan:
@matthewhoffman @xylar - This is ready to review. Thanks :)

xylar:
@matthewhoffman or @hollyhan, does either of you have an example CISM file handy that I could use to test this on?

matthewhoffman:
@hollyhan & @xylar , this looks great - thanks both of you for working through the details of updating this.

@trhille , do you have a simple regular grid GIS or AIS dataset that could be used for testing this regridding script?  We call it 'cism' format here, but we don't actually use CISM files anymore.

xylar:
@matthewhoffman and @trhille, @hollyhan sent me a file to work with so I'm set.

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
```diff
@@ -48,18 +64,20 @@
...
-# ===============================================
+# ===================================
 
 fin = netCDF4.Dataset(options.cismFile, 'r')
 fout = netCDF4.Dataset(options.scripFile, 'w')  # This will clobber existing files
 
 # Get info from input file
-x1 = fin.variables['x1'][:]
-y1 = fin.variables['y1'][:]
+x1 = fin.variables['x'][:]
```
matthewhoffman, line 79: the `x1` and `y1` dimension variable names were used in CISM, and for awhile we primarily were using CISM files as source datasets.  At this point we don't do that anymore and instead use a range of regular ice sheet data source files.  I believe they typically have dimension variable names of `x` and `y`, so we could just change that everywhere in this script.  We could also go and change the filename as well, e.g. `create_SCRIP_file_from_planar_rectangular_grid.py`

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
```diff
@@ -17,26 +22,37 @@
...
 #projections['gis-bamber'] = pyproj.Proj('+proj=stere +lat_ts=71.0 +lat_0=90 +lon_0=321.0 +k_0=1.0 +x_0=800000.0 +y_0=3400000.0 +geoidgrids=./egm08_25.gtx')
-projections['gis-bamber'] = pyproj.Proj('+proj=stere +lat_ts=71.0 +lat_0=90 +lon_0=321.0 +k_0=1.0 +x_0=800000.0 +y_0=3400000.0 +ellps=WGS84')  # This version ignores the vertical datum shift, which should be a very small error for horizontal-only positions
+#projections['gis-bamber'] = pyproj.Proj('+proj=stere +lat_ts=71.0 +lat_0=90 +lon_0=321.0 +k_0=1.0 +x_0=800000.0 +y_0=3400000.0 +ellps=WGS84')  # This version ignores the vertical datum shift, which should be a very small error for horizontal-only positions
+
+# GIMP projection: This is also polar stereographic but with different standard parallel and using the WGS84 ellipsoid.
+#projections['gis-gimp'] = pyproj.Proj('+proj=stere +lat_ts=70.0 +lat_0=90 +lon_0=315.0 +k_0=1.0 +x_0=0.0 +y_0=0.0 +ellps=WGS84')
+
+
+
+# HH: the above dictionary for projections are commented out, and modified PROJ strings (compatible with Pyproj v3) are added below
```
matthewhoffman, line 32: You can go ahead and delete the commented version that is no longer supported.  Thanks for updating to the new pyproj requirements.

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
```diff
@@ -2,11 +2,16 @@
...
 
+# This script has been updated by Holly Han (Postdoc in Matt Hoffman's group in 2021) to be compatible with Python 3 and Pyproj 2 versions.
+# The previous script used the class "transform" from Pyproj v1, which is deprecated in the new versions.
+# To see what have been modified, see comments that start with "# HH".  (Holly Han, September 1st, 2021).
+
 import sys
 import netCDF4
 import numpy as np
 from optparse import OptionParser
 import pyproj
```
matthewhoffman, line 13: Is the plain import of `pyproj` still needed?  If not, go ahead and delete it.

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
```diff
@@ -122,23 +166,33 @@
...
-#plt.plot(grid_corner_lon[i, 0], grid_corner_lat[i, 0], 'kx')
-#plt.plot(grid_corner_lon[i, 1], grid_corner_lat[i, 1], 'bx')
-#plt.plot(grid_corner_lon[i, 2], grid_corner_lat[i, 2], 'cx')
-#plt.plot(grid_corner_lon[i, 3], grid_corner_lat[i, 3], 'gx')
-#plt.show()
-##plot all points
-#plt.plot(grid_center_lon[:], grid_center_lat[:], 'bo')
-#plt.plot(grid_corner_lon[:], grid_corner_lat[:], 'g.')
-#plt.show()
+import matplotlib.pyplot as plt
```
matthewhoffman, line 182: @hollyhan , do you think it is useful to leave this uncommented?  Or would it be better to re-comment it?

matthewhoffman (changes requested):
@hollyhan & @xylar , thanks for working on updating this script.  I made a few comments below.  I'm not sure if we will use this enough to justify a lot of work on small details, but it's also a good first PR exercise!

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
xylar: I personally think you should add an optional `--plot` flag and call this if `--plot` is present.  But it should be commented out if you don't want to do that.

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
```diff
@@ -122,23 +166,33 @@
...
+##HH
+if int(options.gridRank) == 1:
+    print('grid dims is nx*xy')
+    grid_dims[:] = (nx * ny) # for RANK 1 data
+elif int(options.gridRank) == 2:
+    print('grid dims is nx,xy')
+    grid_dims[:] = [nx , ny]
+
+
+#grid_dims[:] = [nx , ny] # HH: for RANK 2 data
```
xylar, line 179: I don't think this comment is needed.  It is clear from the if/else above.

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
```diff
@@ -122,23 +166,33 @@
...
 grid_imask[:] = 1  # For now, assume we don't want to mask anything out - but eventually may want to exclude certain cells from the input mesh during interpolation
-grid_dims[:] = (nx * ny)
 
+##HH
+if int(options.gridRank) == 1:
+    print('grid dims is nx*xy')
+    grid_dims[:] = (nx * ny) # for RANK 1 data
+elif int(options.gridRank) == 2:
+    print('grid dims is nx,xy')
+    grid_dims[:] = [nx , ny]
```
xylar, line 176: It's always best to have `if`/`elseif`/`else`.  This way, there is a clear error message when a wrong parameter value is given:
```suggestion
if int(options.gridRank) == 1:
    print('grid dims is nx*xy')
    grid_dims[:] = (nx * ny) # for RANK 1 data
elif int(options.gridRank) == 2:
    print('grid dims is nx,xy')
    grid_dims[:] = [nx , ny]
else:
    raise ValueError(f'gridRank must be 1 or 2 but {options.gridRank} was given.') 
```

xylar: Oh, I see, you do this above...

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
```diff
@@ -69,7 +87,18 @@
...
+##HH
+if int(options.gridRank) == 1:
+    print('grid rank is 1')
+    fout.createDimension("grid_rank", 1)
+elif int(options.gridRank) == 2:
+    print('grid rank is 2')
+    fout.createDimension("grid_rank", 2)
+else:
+    print('grid rank value is invalid: valid options are 1 or 2. Please try it again.')
+    exit()
```
xylar, line 100: A more typical way to do this would be:
```suggestion
else:
    raise ValueError(f'grid rank value is invalid: valid options are 1 or 2 but {options.gridRank} was given.') 
```

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
xylar: I would strongly support the name change if it isn't being used for CISM grids.  You could also add `-x` and `-y` flags that default to `x` and `y` if you thought you would ever need the `x1` and `y1` versions again.

xylar:
We include the [`create_SCRIP_file_from_MPAS_mesh.py`](https://github.com/MPAS-Dev/MPAS-Tools/blob/master/mesh_tools/create_SCRIP_files/create_SCRIP_file_from_MPAS_mesh.py) in the `mpas_tools` conda package.  Would it be helpful to include this script, too?  If so, it should be added below here:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/conda_package/setup.py#L64

@matthewhoffman, what do you think?

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
hollyhan: I now remember that I commented this section out to do quick testing of the code without plotting. I could revise it and add an optional --plot flag as Xylar suggests.

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
hollyhan: Oh, very cool. Got it! Thanks, Xylar :)

matthewhoffman:
@xylar , now that you mention it, yes it would be useful to include this as well in mpas-tools package.  The use case I am thinking of is when we need to do conservative remapping from ice sheet source datasets to MALI meshes in compass.  We have historically just done these as one-offs but it would be valuable to have the ability within compass.  Should we add that package change to this PR in an additional commit?  (after the script name change)

xylar:
> Should we add that package change to this PR in an additional commit? (after the script name change)

Yes, I think that makes sense.  No need to keep it separate.

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
hollyhan: I've changed the variable names 'x1' and 'y1' to 'x' and 'y'

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
hollyhan: Now that I'm going through the revision, I actually think it would be useful to leave this whole block uncommented because it's a useful checking tool for output scrip files. I realize I commented the block out because I was doing a number of quick tests to see if the code runs well after some edits, but then I never remembered to uncommented it after.

xylar: Hmm, in your version of the code (in green) the plotting isn't commented out.  So I'm guessing you uncommented the plotting to get some plots to see what is going on.

Are you suggesting just switching back to the commented-out version?  I think that was @matthewhoffman's suggestion.  I still think it would be a better practice to add flags to the arguments rather than expecting users to modify the script.  For one thing, modifying the script within a conda environment is both hard and not a good practice. So I still advocate adding a command-line flag `--plot` and putting an `if arg.plot:` statement around this plotting code (except the `import`, which needs to go to the top of the file).

hollyhan: Okay, I'm definitely getting my memory mixed up! Okay, I will include a command-line plot! I've also moved the import matplotlib part to the top of the file.

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
```diff
@@ -16,60 +17,70 @@
...
+print ("== Gathering information.  (Invoke with --help for more details. All arguments are optional)")
 parser = OptionParser()
 parser.description = "This script takes an MPAS grid file and generates a SCRIP grid file."
-parser.add_option("-c", "--cism", dest="cismFile", help="CISM grid file name used as input.", default="cism.nc", metavar="FILENAME")
+parser.add_option("-i", "--input", dest="inputFile", help="input grid file name used as input.", default="input.nc", metavar="FILENAME")
 parser.add_option("-s", "--scrip", dest="scripFile", help="SCRIP grid file to output.", default="scrip.nc", metavar="FILENAME")
-parser.add_option("-p", "--proj", dest="projection", help="projection used by the CISM file. Valid options are:"  + str(projections.keys()), metavar="PROJ")
+parser.add_option("-p", "--proj", dest="projection", help="projection used by the input data file. Valid options are:"  + str(projections.keys()), metavar="PROJ")
+parser.add_option("-r", "--rank", dest="gridRank", help="desired rank of the output SCRIP grid data")
+parser.add_option("--plot", dest="plot", help="destination grid points are plotted if 'True'. Default is 'False'", default=False)
```
xylar, line 41: There's a slightly cleaner way to handle this:
https://docs.python.org/3/library/optparse.html#handling-boolean-flag-options

```suggestion
parser.add_option("--plot", dest="plot', action="store_true", help="If this flag is used, destination grid points are plotted")
```

xylar:
> I'm not sure how to start conversation on the new page 

@hollyhan, to start a conversation about a part of the code, you go to "Files changed" at the top of the page, scroll down to the code changes, move your mouse over to the left margin (before the line numbers) so you see a blue plus sign, and select the line(s) you want to discuss.  Then, a small conversation window will pop up.  You can use the "Insert a suggestion" button (a page with a +/- on it) to make suggestions about changes to the code like I've been doing.

Some other handy code formatting tips for GitHub are here:
https://guides.github.com/features/mastering-markdown/
These also work (pretty well) on Slack.

xylar:
@hollyhan, I think your last commit added a new file named `create_SCRIP_file_from_planar_rectangular_grid.py` but didn't delete the one with the old name.  Could you do:
```
git rm mesh_tools/create_SCRIP_files/create_SCRIP_file_from_CISM_mesh.py
git commit
git push ...
```
That should take care of it.

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_planar_rectangular_grid.py
```diff
@@ -0,0 +1,172 @@
...
+    plt.plot(grid_corner_lon[i, 0], grid_corner_lat[i, 0], 'kx')
+    plt.plot(grid_corner_lon[i, 1], grid_corner_lat[i, 1], 'bx')
+    plt.plot(grid_corner_lon[i, 2], grid_corner_lat[i, 2], 'cx')
+    plt.plot(grid_corner_lon[i, 3], grid_corner_lat[i, 3], 'gx')
+    plt.show()
+
+    #plot all points
+    plt.plot(grid_center_lon[:], grid_center_lat[:], 'bo')
+    plt.plot(grid_corner_lon[:], grid_corner_lat[:], 'g.')
+    plt.show()
```
xylar, line 168: This is super picky since this is just for debugging but maybe plot the two figures and then show them both:
```suggestion
    plt.figure(1)
    plt.plot(grid_center_lon[i], grid_center_lat[i], 'o')
    plt.plot(grid_corner_lon[i, 0], grid_corner_lat[i, 0], 'kx')
    plt.plot(grid_corner_lon[i, 1], grid_corner_lat[i, 1], 'bx')
    plt.plot(grid_corner_lon[i, 2], grid_corner_lat[i, 2], 'cx')
    plt.plot(grid_corner_lon[i, 3], grid_corner_lat[i, 3], 'gx')

    #plot all points
    plt.figure(2)
    plt.plot(grid_center_lon[:], grid_center_lat[:], 'bo')
    plt.plot(grid_corner_lon[:], grid_corner_lat[:], 'g.')
    plt.show()
```

hollyhan:
Hi @xylar , yes! Please go ahead and push the change to the branch - you won't be stepping on anyone's toes! Thanks for checking with me.

xylar:
@hollyhan, the changes I made are on my work laptop so I'll push them on Monday.  Thanks for the latest update.

xylar:
@hollyhan, let me know if you've run the extra steps (`ESMF_RegridWweightGen` or `ncremap`) that @matthewhoffman mentioned and if they were successful.  If so, I'll merge.

xylar:
Thanks again @hollyhan and @matthewhoffman!

PR #448, opened 2021-11-18
Fix units in converting smb to sfcMassBal in landice interpolation script

trhille:
Was improperly using ice density instead of water density when
converting mm w.e. yr^{-1} to kg m^{-2} s^{-1}.


matthewhoffman (approved):
Approved by inspection.  Thanks for catching this, @trhille .

PR #449, opened 2021-11-26
Add keyword arguments to `check_call`

xylar:
The keyword arguments are passed along to `subprocess.Popen`. This is useful if a given call needs to do things like pass a new environment for `Popen` to use.

This merge also updates the version number to 0.12.0


xylar:
## Testing

I have run the `compass` nightly test suite with these changes and it ran successfully.

PR #450, opened 2022-01-14
Fix mpas mesh tools

hollyhan:
Hi Xylar and Matt, 

In discussing with @matthewhoffman, we found the need to make some changes to two mesh tool scripts so that the mesh/scripfiles generated from these tools have positive longitude convention [0 2pi], which is required by MPAS mesh spec (https://mpas-dev.github.io/files/documents/MPAS-MeshSpec.pdf)

Your review on these changes would be very appreciated. 
Thank you!
Holly


conda_package/mpas_tools/scrip/from_mpas.py
```diff
@@ -11,15 +11,12 @@
...
     ----------
     mpasFile : str
         The path to a NetCDF file with the MPAS mesh
-
     scripFile : str
         The path to the output SCRIP file
-
     useLandIceMask : bool
         Whether to use the landIceMask field for masking
     """
```
xylar, line 22: I would prefer to leave the docstring spacing as it was.  I think it's more readable with the extra blank lines.

conda_package/mpas_tools/scrip/from_mpas.py
```diff
@@ -47,10 +44,13 @@ def scrip_from_mpas(mpasFile, scripFile, useLandIceMask=False):
...
 
+    # change the longitude convention to use positive values [0 2pi]
+    lonCell[lonCell<0.0] += 2.0*np.pi
+    lonVertex[lonVertex<0.0] += 2.0*np.pi
+
     if sphereRadius <= 0:
-        print(" -- WARNING: conservative remapping is NOT possible when "
-              "'sphereRadius' <= 0 because 'grid_area' field will be infinite "
-              "(from division by 0)")
+       sphereRadius =  6.37122e6
```
xylar, line 52: I would prefer to get this from `mpas_tools.cime.constants` rather than hard coding it:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/conda_package/mpas_tools/cime/constants.py#L6

We will need to add a new import at the top:
```
from mpas_tools.cime.constants import constants
```
and then make this change:
```suggestion
       sphereRadius =  constants['SHR_CONST_REARTH']
```

conda_package/mpas_tools/scrip/from_mpas.py
```diff
@@ -47,10 +44,13 @@ def scrip_from_mpas(mpasFile, scripFile, useLandIceMask=False):
...
+       print(" -- WARNING: sphereRadius<0 so setting sphereRadius= 6.37122e6")
```
xylar, line 53:
```suggestion
       print(f" -- WARNING: sphereRadius<0 so setting sphereRadius= {constants['SHR_CONST_REARTH']}")
```

conda_package/mpas_tools/scrip/from_mpas.py
```diff
@@ -168,3 +168,6 @@ def main():
         options.landiceMasks = False
 
     scrip_from_mpas(options.mpasFile, options.scripFile, options.landiceMasks)
+
+if __name__ == "__main__":
+     main()
```
xylar, line 173: This file is not meant to be run as a script.  It is an entry point in the conda package.  So I would prefer not to add this.

xylar: I'm happy to discuss this futher.

conda_package/mpas_tools/scrip/from_mpas.py
```diff
@@ -47,10 +44,13 @@ def scrip_from_mpas(mpasFile, scripFile, useLandIceMask=False):
     sphereRadius = float(fin.sphere_radius)
     on_a_sphere = str(fin.on_a_sphere)
 
+    # change the longitude convention to use positive values [0 2pi]
+    lonCell[lonCell<0.0] += 2.0*np.pi
+    lonVertex[lonVertex<0.0] += 2.0*np.pi
+
```
xylar, line 50: Here and in the corresponding edit below, should there be a check to see if `lonCell` and `lonVertex` are `> 2.0*np.pi`?  I believe the more compact way to ensure that these are really in the intended range would be:
```suggestion
    # change the longitude convention to use positive values [0 2pi]
    lonCell = numpy.mod(lonCell, 2.0*np.pi)
    lonVertex = numpy.mod(lonVertex, 2.0*np.pi)
```
I verified that this works for negative longitudes:
```
$ python
>>> import numpy
>>> numpy.mod(-numpy.pi, 2*numpy.pi)
3.141592653589793
```

xylar (commented):
Thanks for the PR!  This is a great thing to fix.  I have a few suggestions that we can discuss when we meet next, or here if you prefer.

conda_package/mpas_tools/scrip/from_mpas.py
xylar: I changed my mind.  If this is helpful for your debugging, it's fine.  In general, I just don't want folks running this file as a script, it should be run through the entry point:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/conda_package/setup.py#L89

matthewhoffman (commented):
Thanks for making this PR, @hollyhan .  It will be worth it in the long run to have taken the time to fix these issues before we forget about them.

I looked over the changes, and I don't have anything to add beyond what @xylar identified.  All of his suggestions are good ones.

conda_package/mpas_tools/scrip/from_mpas.py
hollyhan: Right, my bad for unnecessarily changing those lines!

conda_package/mpas_tools/scrip/from_mpas.py
hollyhan: I see. I was initially following the design document for MPAS-Ocean-SAL (Self Attraction and Loading, https://docs.google.com/document/d/1t2YnII83zK2ajlWAalK5sA8HoE0dglrB7AEiV6OQDvM/edit), which hints users to run the script on the command line to generate a scripfile for the MPAS mesh.  I assume the design document then has some incorrect instruction on how to use the script. We could talk about this tomorrow and decide on what we want to do about this proposed change.

conda_package/mpas_tools/scrip/from_mpas.py
```diff
@@ -11,15 +11,15 @@
...
     ----------
     mpasFile : str
         The path to a NetCDF file with the MPAS mesh
-
+        
     scripFile : str
         The path to the output SCRIP file
-
+        
     useLandIceMask : bool
```
xylar, line 23: @hollyhan, unfortunately in the process of putting these blank lines back they ended up with trailing white space, something we try to avoid.  Could you take those out?  Thanks!

mesh_tools/planar_grid_transformations/set_lat_lon_fields_in_planar_grid.py
```diff
@@ -77,13 +77,18 @@
 
 print("Input file xCell min/max values:", xCell[:].min(), xCell[:].max())
 print("Input file yCell min/max values:", yCell[:].min(), yCell[:].max())
-
+
```
xylar, line 80: Here is another bit of white space to take out.  Thanks again!

xylar (commented):
Other than the two super nit-picky comments I just put in, this looks great!  I just want to run a quick test, then I'll approve.

conda_package/mpas_tools/scrip/from_mpas.py
```diff
@@ -47,10 +47,12 @@ def scrip_from_mpas(mpasFile, scripFile, useLandIceMask=False):
     sphereRadius = float(fin.sphere_radius)
     on_a_sphere = str(fin.on_a_sphere)
 
+    # change the longitude convention to use positive values [0 2pi]
+    lonCell = numpy.mod(lonCell, 2.0*np.pi)
+    lonVertex = numpy.mod(lonVertex, 2.0*np.pi)
```
xylar, line 52: Just as a side comment which I totally missed when I looked at this before:

This function is for creating a SCRIP file from an MPAS mesh.  I think MPAS will care if the longitudes are between 0 and 2*pi but the scrip file should work fine regardless of which longitude convention is being used.  In other words, if someone is supplying a bad MPAS mesh to this function, it might be pretty late to fix the longitude.

@hollyhan, did you find that you weren't able to create scrip files from some meshes because of this?

xylar:
Hmm, my tests weren't successful.  I did this to create a test conda environment:
```
conda create -y -n test_mesh_tools --file conda_package/dev-spec.txt 
conda activate test_mesh_tools
python -m pip install -e conda_package/
```
I verified that it worked (that it has the expected entry point):
```
scrip_from_mpas --help
```
But when I ran it on a file we use in COMPASS, it didn't work:
```
Traceback (most recent call last):
  File "/home/xylar/miniconda3/envs/test_mesh_tools/bin/scrip_from_mpas", line 33, in <module>
    sys.exit(load_entry_point('mpas-tools', 'console_scripts', 'scrip_from_mpas')())
  File "/home/xylar/code/MPAS-Tools/fix_MPAS_mesh_tools/conda_package/mpas_tools/scrip/from_mpas.py", line 172, in main
    scrip_from_mpas(options.mpasFile, options.scripFile, options.landiceMasks)
  File "/home/xylar/code/MPAS-Tools/fix_MPAS_mesh_tools/conda_package/mpas_tools/scrip/from_mpas.py", line 51, in scrip_from_mpas
    lonCell = numpy.mod(lonCell, 2.0*np.pi)
NameError: name 'numpy' is not defined
```
It needs to be `np` instead of `numpy`.  That was my mistake in my original suggestion, so I'm sorry about that.

With this fix, I was able to generate a script file that was identical to the one generated by the old `scrip_from_mpas`.

Similarly, when I tried to run `set_lat_lon_fields_in_planar_grid.py`, I see:
```
Traceback (most recent call last):
  File "/home/xylar/miniconda3/envs/test_mesh_tools/bin/set_lat_lon_fields_in_planar_grid.py", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/xylar/code/MPAS-Tools/fix_MPAS_mesh_tools/conda_package/mesh_tools/planar_grid_transformations/set_lat_lon_fields_in_planar_grid.py", line 88, in <module>
    lonCell = numpy.mod(lonCell, 2.0*np.pi)
NameError: name 'numpy' is not defined
```
This time, `numpy` isn't getting imported at all, so you need to change `numpy` to `np` again and add `import numpy as np` at the top of the file.

When I made this change, I was able to add latitude and longitude to the mesh for the compass dome test case:
```
set_lat_lon_fields_in_planar_grid.py -f mpas_grid.nc -p ais-bedmap2
```
Things looked as expected, though this wasn't a very rigorous test because longitudes are all between 0 and pi/2...

xylar:
@hollyhan, I retested tested with the changes you made.  I had to go to a little extra trouble to get the C++ tools without actually building the conda package:
```
conda create -y -n test_mpas_tools python=3.9 mpas_tools=0.12.0
conda activate test_mpas_tools
```
That creates an environment with the latest release of `mpas_tools`, including the C++ tools like `MpasCellCuller.x`.

Then I install over top of it with the changes you made:
```
python -m pip install -e conda_package/
```
This is a little risky in general for reasons I'm happy to explain (dependencies may have changed since the last release, for example) but it's good enough for testing.

Then, I created a mesh to test with:
```
# create a planar mesh
planar_hex --nx 30 --ny 34 --dc 2000.0 --npx --npy -o test.nc
# cull the periodic boundaries
MpasCellCuller.x test.nc culled.nc
# center the mesh at (0,0)
translate_planar_grid -f culled.nc -c
# add lat/lon
set_lat_lon_fields_in_planar_grid.py -f culled.nc --proj ais-bedmap2
# make sure lonCell are all positive:
ncdump -v lonCell culled.nc
```
This all worked great.

Then, I tested the SCRIP file generation:
```
$ scrip_from_mpas -m culled.nc -s scrip.nc
== Gathering information.  (Invoke with --help for more details. All arguments are optional)
 -- Landice Masks are disabled

Traceback (most recent call last):
  File "/home/xylar/miniconda3/envs/test_mesh_tools/bin/scrip_from_mpas", line 33, in <module>
    sys.exit(load_entry_point('mpas-tools', 'console_scripts', 'scrip_from_mpas')())
  File "/home/xylar/code/MPAS-Tools/fix_MPAS_mesh_tools/conda_package/mpas_tools/scrip/from_mpas.py", line 173, in main
    scrip_from_mpas(options.mpasFile, options.scripFile, options.landiceMasks)
  File "/home/xylar/code/MPAS-Tools/fix_MPAS_mesh_tools/conda_package/mpas_tools/scrip/from_mpas.py", line 54, in scrip_from_mpas
    sphereRadius =  constants['SHR_CONST_REARTH']
NameError: name 'constants' is not defined
```
This was a surprise because I thought this worked when I tested yesterday.  But it seems like we're missing an import here:
```
from mpas_tools.cime.constants import constants
```

I added this myself and things worked as expected.

hollyhan:
Thanks for going through the testing and explaining the steps in detail, @xylar ! Please let me know if there's anything I should do further to get this PR closed.

conda_package/mpas_tools/scrip/from_mpas.py
hollyhan: @xylar 

That's a good point. I was able to create scripfiles using this code, but the main issue of having a scripfile that uses the negative longitude convention for us came up when I was using ESMF to generate remapping weights `ESMF_RegridWeightGen` to remap data from the MALI mesh (which used the negative long. convention) onto the global Gaussian (SLM) grid (which used the positive long. convention).  The generated weights were incorrect, and we found out that EMSF doesn't support different longitude conventions for the SCRIP format. @matthewhoffman - I'm tagging you in case  you have something to add to this.

xylar: Oh, that's a big surprise to me. I've used MPAS meshes (0 to 2*pi) as inputs and lon/lat grids with longitudes between -180 and 180 degrees all the time without trouble.  But I might be converting MPAS longitudes to the range expected by ESMF.  Let me check.

xylar: No, it doesn't look like I do anything to `lonCell` and this has always worked fine for me:
https://github.com/MPAS-Dev/pyremap/blob/master/pyremap/descriptor.py#L228-L229
And here's verifying that I use -180 to 180 as the bounds for my lat/lon meshes:
https://github.com/MPAS-Dev/pyremap/blob/master/pyremap/descriptor.py#L38-L39

xylar:
@hollyhan, as I said above, `from mpas_tools.cime.constants import constants` needs to be added to `mpas_tools/scrip/from_mpas.py`.  I did this in my local testing but I didn't modify this branch.

Then, you and I need to work together on rebasing this branch to clean thing up into just 2 or 3 commits.

conda_package/mpas_tools/scrip/from_mpas.py
xylar: @matthewhoffman, no, I'm quite certain the source MPAS mesh is using 0 to 2*pi radians and the destination mesh is using -180 to 180 degrees.  It could be that mixing radians and degrees means that one gets converted to the other and, in the process, they get converted to the same range.  But I have also never seen any warnings from ESMF that you need to have these values in a consistent range.  In fact, I believe I have also always used **inconsistent** (-180 to 180 degree) ranges for all my destination grids, even the Antarctic and Arctic projection grids.

matthewhoffman: Hmm, that is troubling.  That issue was what motivated opening the PR in the first place, but if it not actually a problem, it suggests there was something else unknown going on with workflow.  Based on your experience, do you think we should *not* be making the change for the longitude convention in this PR?  Or should we retain them under the assumption that they do no harm but possibly are helping us for an unknown reason?

xylar: @matthewhoffman, hmm.  I still think that any MPAS mesh that needs this fix isn't being constructed properly.  It would have maybe made more sense to me to have a sanity check to make sure `lon*` is between 0 and 2*pi, rather than a fix to ensure it is.  If you are passing what you think is an MPAS mesh to this routine and the longitude isn't following the MPAS convention, you should probably fix it somewhere upstream in your workflow.  What to you think?

matthewhoffman: I'm a little confused because it looks like we are now discussing an outdated change, so maybe the two of you decided to remove this adjustment.  

In any case, @xylar , I agree that it would be best to fix these issues at the most upstream place.  For a MALI mesh, that's in the set_lat_lon... script, the correction for which is also in this PR.  I'll point out that MALI itself does not use the latCell/lonCell (and edge & vertex versions), so we only add those fields via that script for regridding purposes.

Regarding this outdated change to from_mpas.py, I agree with your argument we don't necessarily want to be modifying the longitude at this point.  However, it might still be a good idea to have an error or a warning if the longitude range is outside of the expected 0-2pi range.

xylar:
@hollyhan, would you be willing to change `from_mpas.py` to check that `lonCell` and `lonVertex` are in the desired range (0, 2*pi) rather than modifying them?  See https://github.com/MPAS-Dev/MPAS-Tools/pull/450#discussion_r796125539

Then, I'll work with you to rebase the PR into 2 or 3 commits as I mentioned last week.

hollyhan:
@xylar  Yes, I'll work on the check for the range of lonCell and lonVertex before we meet on Wednesday.

xylar:
@hollyhan, so we still missed a step in `from_mpas.py`.  When I test, I see:
```
ed.nc -s scrip.nc
== Gathering information.  (Invoke with --help for more details. All arguments are optional)
 -- Landice Masks are disabled

Traceback (most recent call last):
  File "/home/xylar/miniconda3/envs/test/bin/scrip_from_mpas", line 33, in <module>
    sys.exit(load_entry_point('mpas-tools', 'console_scripts', 'scrip_from_mpas')())
  File "/home/xylar/code/MPAS-Tools/fix_MPAS_mesh_tools/conda_package/mpas_tools/scrip/from_mpas.py", line 177, in main
    scrip_from_mpas(options.mpasFile, options.scripFile, options.landiceMasks)
  File "/home/xylar/code/MPAS-Tools/fix_MPAS_mesh_tools/conda_package/mpas_tools/scrip/from_mpas.py", line 58, in scrip_from_mpas
    sphereRadius =  constants['SHR_CONST_REARTH']
NameError: name 'constants' is not defined
```
Could you add the missing import:
```
from mpas_tools.cime.constants import constants
```
 and do the following?
```
git commit --amend from_mpas.py
git push --force hollyhan/MPAS-Tools fix_MPAS_mesh_tools
```

hollyhan:
@xylar I've added the missing import and pushed the branch

xylar (approved):
Yep, @hollyhan, that did the trick.

I have tested this to my satisfaction.  Thanks so much for the hard work.  It seemed like a simple task but I managed to make it complicated!

hollyhan:
I tested the new `from_mpas.py` with the thwaites mesh that uses the longitude convention of [-pi pi].
The test results shows the following error, as we expect:

```scrip_from_mpas -m thwaites.4km.210608.nc -s scrip.nc 
== Gathering information.  (Invoke with --help for more details. All arguments are optional)
 -- Landice Masks are disabled

Traceback (most recent call last):
  File "/Users/hollyhan/miniconda3/envs/test_mpas_tools/bin/scrip_from_mpas", line 33, in <module>
    sys.exit(load_entry_point('mpas-tools', 'console_scripts', 'scrip_from_mpas')())
  File "/Users/hollyhan/Desktop/E3SM-Project/MPAS_Tools/fix_MPAS_mesh_tools/conda_package/mpas_tools/scrip/from_mpas.py", line 177, in main
    scrip_from_mpas(options.mpasFile, options.scripFile, options.landiceMasks)
  File "/Users/hollyhan/Desktop/E3SM-Project/MPAS_Tools/fix_MPAS_mesh_tools/conda_package/mpas_tools/scrip/from_mpas.py", line 52, in scrip_from_mpas
    raise ValueError("lonCell is not in the desired range (0, 2pi)")
ValueError: lonCell is not in the desired range (0, 2pi)```

**Review by matthewhoffman — APPROVED** [review]
@hollyhan , these changes make sense based on where the discussion ended up.  If you haven't already done so, it's probably a good idea to rerun your workflow of adding lat/lon to a mesh file, running the remap tool, and making sure the map file works properly for regridding inside your SLM code in MALI.

**Comment by xylar** [comment]
@hollyhan, I'll merge this as soon as you have re-tested your workflow and let us know that it worked as expected.

**Comment by hollyhan** [comment]
Hi have tested the entire workflow:

**1. Try creating a scripfile from a thwaites mesh that uses undesired longitude convention:** 
```
$ scrip_from_mpas -m thwaites.4km.220203.nc -s thwaites.4km.220203.scripfile.nc 

== Gathering information.  (Invoke with --help for more details. All arguments are optional)
 -- Landice Masks are disabled

Traceback (most recent call last):
  File "/Users/hollyhan/miniconda3/envs/test_mpas_tools/bin/scrip_from_mpas", line 33, in <module>
    sys.exit(load_entry_point('mpas-tools', 'console_scripts', 'scrip_from_mpas')())
  File "/Users/hollyhan/Desktop/E3SM-Project/MPAS_Tools/fix_MPAS_mesh_tools/conda_package/mpas_tools/scrip/from_mpas.py", line 177, in main
    scrip_from_mpas(options.mpasFile, options.scripFile, options.landiceMasks)
  File "/Users/hollyhan/Desktop/E3SM-Project/MPAS_Tools/fix_MPAS_mesh_tools/conda_package/mpas_tools/scrip/from_mpas.py", line 52, in scrip_from_mpas
    raise ValueError("lonCell is not in the desired range (0, 2pi)")
ValueError: lonCell is not in the desired range (0, 2pi)
```
Error appears as expected. 

**2. Change the longitude convention with the new changes made in this PR**
```
$ set_lat_lon_fields_in_planar_grid.py -f thwaites.4km.220203.nc --proj ais-bedmap2 
== Gathering information.  (Invoke with --help for more details. All arguments are optional)

Using ais-bedmap2 projection, defined as: +proj=stere +lat_0=-90 +lat_ts=-71 +lon_0=0 +x_0=0 +y_0=0 +ellps=WGS84 +units=m +no_defs
Input file xCell min/max values: -1623967.0094751006 -1073016.3730225994
Input file yCell min/max values: -828184.9435592396 -161464.9911215813
/Users/hollyhan/Desktop/E3SM-Project/MPAS_Tools/fix_MPAS_mesh_tools/conda_package/mesh_tools/planar_grid_transformations/set_lat_lon_fields_in_planar_grid.py:84: DeprecationWarning: This function is deprecated. See: https://pyproj4.github.io/pyproj/stable/gotchas.html#upgrading-to-pyproj-2-from-pyproj-1
  lonCell, latCell = pyproj.transform(projections[options.projection], projections['latlon'], xCell[:], yCell[:], radians=True)
/Users/hollyhan/Desktop/E3SM-Project/MPAS_Tools/fix_MPAS_mesh_tools/conda_package/mesh_tools/planar_grid_transformations/set_lat_lon_fields_in_planar_grid.py:85: DeprecationWarning: This function is deprecated. See: https://pyproj4.github.io/pyproj/stable/gotchas.html#upgrading-to-pyproj-2-from-pyproj-1
  lonVertex, latVertex = pyproj.transform(projections[options.projection], projections['latlon'], xVertex[:], yVertex[:], radians=True)
/Users/hollyhan/Desktop/E3SM-Project/MPAS_Tools/fix_MPAS_mesh_tools/conda_package/mesh_tools/planar_grid_transformations/set_lat_lon_fields_in_planar_grid.py:86: DeprecationWarning: This function is deprecated. See: https://pyproj4.github.io/pyproj/stable/gotchas.html#upgrading-to-pyproj-2-from-pyproj-1
  lonEdge, latEdge = pyproj.transform(projections[options.projection], projections['latlon'], xEdge[:], yEdge[:], radians=True)
Calculated latCell min/max values (radians): -1.3943921358761688 -1.3008508250402886
Calculated lonCell min/max values (radians): 4.119929571080232 4.586072104970466
Lat/lon calculations completed.  File has been written.
```
**3. Create a scripfile with the new mesh that uses the desired longitude convention**
```
$ scrip_from_mpas -m thwaites.4km.220203.nc -s thwaites.4km.220203.scripfile.nc 
== Gathering information.  (Invoke with --help for more details. All arguments are optional)
 -- Landice Masks are disabled

 -- WARNING: sphereRadius<0 so setting sphereRadius = 6371220.0
 -- WARNING: 'on_a_sphere' attribute is 'NO', which means that there may be some disagreement regarding area between the planar (source) and spherical (target) mesh
Input latCell min/max values (radians): -1.3943921358761688, -1.3008508250402886
Input lonCell min/max values (radians): 4.119929571080232, 4.586072104970466
Calculated grid_center_lat min/max values (radians): -1.3943921358761688, -1.3008508250402886
Calculated grid_center_lon min/max values (radians): 4.119929571080232, 4.586072104970466
Calculated grid_area min/max values (sq radians): 4.728403336933561e-07, 2.4751188489024916e-06
Creation of SCRIP file is complete.

```

**4. Create a mapfile going from the mpas mesh to the gaussian grid using ESMF weight generator.**
```
$ ESMF_RegridWeightGen -d gauss_latlon512x1024_rank2_n2s_scripfile.nc -s thwaites.4km.220203.scripfile.nc -w mpas_to_grid.220203.nc -i --64bit_offset  --src_regional

 Starting weight generation with these inputs: 
   Source File: thwaites.4km.220203.scripfile.nc
   Destination File: gauss_latlon512x1024_rank2_n2s_scripfile.nc
   Weight File: mpas_to_grid.220203.nc
   Source File is in SCRIP format
   Source Grid is a regional grid
   Source Grid is an unstructured grid
   Use the center coordinates of the source grid to do the regrid
   Destination File is in SCRIP format
   Destination Grid is a global grid
   Destination Grid is a logically rectangular grid
   Use the center coordinates of the destination grid to do the regrid
   Regrid Method: bilinear
   Pole option: NONE
   Ignore unmapped destination points
   Output weight file in 64bit offset NetCDF file format
   Line Type: cartesian
   Norm Type: dstarea
   Extrap. Method: none
```
**5. Interpolate ice thickness from the thwaites mesh to the global gaussian grid using ncremap**
```
$ ncremap -m mpas_to_grid.220203.nc -i thwaites.4km.220203.nc -o interp_mpas_to_grid_results_220203.nc -v thickness

```
The resulting interpolated thwaites ice thickness is correct.

hollyhan:
Screen Shot 2022-02-03 at 12 44 07 PM

xylar:
@hollyhan, thanks for documenting your testing carefully.  It looks good!

matthewhoffman:
Yes, thanks @hollyhan for documenting the workflow!  I'm sure it will be useful to have this for reference in the PR history.

xylar:
@matthewhoffman, @hollyhan and @trhille, I'll do a release of MPAS-Tools soon and update the version in `compass` so you can make use of this

PR #456, opened 2022-04-01
Small fix to python mask creation

xylar:
This fix avoids a deprecation warning from Shapely but explicitly getting the list of geometries in a `MultiPolygon` object with the `.geoms` property.


xylar:
## Testing

I have run the QU240 mesh in compass through the `files_for_e3sm` test case, which uses the python masking.  With an `mpas_tools` package built from this branch, I no longer see the deprecation warnings that I was seeing in previous runs.

PR #457, opened 2022-04-01
Remove an unnecessary comment line

hollyhan:
This commit removes the comment line that was not removed before the previous merge.


xylar (approved):
Yep, I agree.  We don't want those kinds of comments.  You can use the "blame" button on GitHub to find out who added or changed a given line.

PR #460, opened 2022-04-16
Add a custom `MpasConfigParser`

xylar:
This is a combination of the config parser from compass and MPAS-Analysis.

The documentation has been updated to include this feature.

### DIFF
--- a/conda_package/mpas_tools/config.py
+++ b/conda_package/mpas_tools/config.py
@@ -15,6 +15,17 @@ class MpasConfigParser:
     of the source of different config options and allows the "user" config
     options to always take precedence over other config options (even if they
     are added later).
+
+    Attributes
+    ----------
+    combined : {None, configparser.ConfigParser}
+        The combined config options
+
+    combined_comments : {None, dict}
+        The combined comments associated with sections and options
+
+    sources : {None, dict}
+        The source of each section or option
     """
 
     _np_allowed = dict(linspace=np.linspace, xrange=range,
@@ -29,9 +40,9 @@ class MpasConfigParser:
         self._configs = dict()
         self._user_config = dict()
         self._comments = dict()
-        self._combined = None
-        self._combined_comments = None
-        self._sources = None
+        self.combined = None
+        self.combined_comments = None
+        self.sources = None
 
     def add_user_config(self, filename):
         """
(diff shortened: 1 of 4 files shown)


xylar:
## Testing

I have branches for using this in [compass](https://github.com/xylar/compass/tree/use_mpas_config_parser) and [MPAS-Analysis](https://github.com/xylar/MPAS-Analysis/tree/use_mpas_config_parser).

I have successfully run the ocean `nightly` test suite on Anvil with the `MpasConfigParser` using the first branch.  I am in the process of running and debugging MPAS-Analysis using the test suite on Anvil.  I will update this comment as soon as those tests are passing and the resulting config file looks as expected.

xylar:
@matthewhoffman, I moved the `CompassConfigParser` over to MPAS-Tools and renamed it `MpasConfigParser`.  I added a bit of functionality (`getexpression()` and `copy()` methods) that has been useful in MPAS-Analysis.  This will let me use the same config parser for both, and get the really great provenance you suggested in MPAS-Analysis, too!

I don't think you need to review this unless you want to.  I just wanted to make you aware of it.

conda_package/mpas_tools/config.py
```diff
@@ -0,0 +1,531 @@
...
+        make sure all paths in the paths, namelists, streams, and executables
+        sections are absolute paths
+        """
+        config = self._combined
+        for section in ['paths', 'namelists', 'streams', 'executables']:
+            if not config.has_section(section):
+                continue
+            for option, value in config.items(section):
+                value = os.path.abspath(value)
+                config.set(section, option, value)
```
xylar, line 459: Shoot, this is something specific to `compass`.  I need to think this through.

xylar: I have solved this issue by making more of the attributes and methods public, and overriding the `combine()` method in the `CompassConfigParser`.  This also disrupts things less in `compass` because we continue to use its version of the config parser while inheriting most of the functionality from here.

xylar:
@matthewhoffman, are you okay with this change, conceptually?  If so, I will go ahead with merging this and making a 0.14.0 release for use in both MPAS-Analysis and compass.  You can help test further once that happens?

matthewhoffman:
@xylar , thanks for pointing this out.  I agree that maintaining a single config parser that works for both COMPAS and MPAS-Analysis is a good idea.  And presumably it is a module that requires only infrequent updating, so having it live in MPAS-Tools is not a limitation to working with it in either package.  I'm happy for you to proceed and to call on me for input as needed.

xylar:
Thanks @matthewhoffman, I appreciate it. I just didn't want to act unilaterally.

PR #461, opened 2022-04-18
Fix string `args` passed to `logging.check_call()`

xylar:
Previously, it assumed that `args` is a list but it now also supports string `args` that allow `shell=True`.  This is useful for joining multiple commands into a single string.


xylar:
## Testing

I used this with this branch of compass:
https://github.com/xylar/compass/tree/add_matrix_build
With this functionality, I am able to run subprocesses with the `shell=True` argument.

PR #463, opened 2022-05-17
Update exodus to mpas conversion script for Cori

trhille:
Add default Seacas path for Cori, udpate load command for exodus to use python3. Set Cori as the default machine, since optimizations and MALI_input_files database are hosted there. Also have stride read from the mpas_cellID.ascii file instead of exodus file, since the definition there is incorrect for these purposes.

### DIFF
--- a/landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
+++ b/landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
@@ -46,11 +46,15 @@ if SEACAS_path == None:
    #sys.path.append('/home/tzhang/Apps/seacas/lib')
    #sys.path.append('/Users/trevorhillebrand/Documents/mpas/seacas/lib/')
    #sys.path.append('/Users/mhoffman/software/seacas/install/lib')
-   sys.path.append('/usr/projects/climate/SHARED_CLIMATE/software/badger/trilinos/2018-12-19/gcc-6.4.0/openmpi-2.1.2/lib')  # path on LANL Badger/Grizzly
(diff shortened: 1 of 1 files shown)


landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
```diff
@@ -46,11 +46,15 @@
    #sys.path.append('/home/tzhang/Apps/seacas/lib')
    #sys.path.append('/Users/trevorhillebrand/Documents/mpas/seacas/lib/')
    #sys.path.append('/Users/mhoffman/software/seacas/install/lib')
-   sys.path.append('/usr/projects/climate/SHARED_CLIMATE/software/badger/trilinos/2018-12-19/gcc-6.4.0/openmpi-2.1.2/lib')  # path on LANL Badger/Grizzly
```
matthewhoffman, line 49: Did you intend to make the Cori path become the default one?  (If I remember correctly, the Badger path no longer points to an existing installation anyway, is that right?)

trhille: Yes, since the optimization and meshes are all on Cori, I thought that was a better default. It might be true that the Badger path is broken. I can check that and update.

trhille: commit [5af2057](https://github.com/MPAS-Dev/MPAS-Tools/pull/463/commits/5af20575f715c75ccc461a5f572927285874573d)

PR #467, opened 2022-06-22
Add an iceMask field to the interpolation script

trhille:
Add an iceMask field to the interpolation script, which can be used
for trimming down fields to the original extent of ice after
interpolating. This is useful when fields must be extrapolated before
using the interpolation script in order to avoid including zeros, NaNs,
or large missing values in the interpolation at the ice edge.

### DIFF
--- a/landice/mesh_tools_li/interpolate_to_mpasli_grid.py
+++ b/landice/mesh_tools_li/interpolate_to_mpasli_grid.py
@@ -649,6 +649,7 @@ fieldInfo = OrderedDict()
 if filetype=='cism':
 
    fieldInfo['thickness'] =     {'InputName':'thk',  'scalefactor':1.0, 'offset':0.0, 'gridType':'x1', 'vertDim':False}
+   fieldInfo['iceMask'] =  {'InputName':'iceMask',  'scalefactor':1.0, 'offset':0.0, 'gridType':'x1', 'vertDim':False}
    if not args.thicknessOnly:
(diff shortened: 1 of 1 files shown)


landice/mesh_tools_li/interpolate_to_mpasli_grid.py
```diff
@@ -649,6 +649,7 @@ def vertical_interp_MPAS_grid(mpas_grid_input_layers, destVertCoord, input_layer
 if filetype=='cism':
 
    fieldInfo['thickness'] =     {'InputName':'thk',  'scalefactor':1.0, 'offset':0.0, 'gridType':'x1', 'vertDim':False}
+   fieldInfo['iceMask'] =  {'InputName':'iceMask',  'scalefactor':1.0, 'offset':0.0, 'gridType':'x1', 'vertDim':False}
```
matthewhoffman, line 652: Did you mean to keep this outside of the thicknessOnly check?

trhille: Although thinking about it more, I think it is placed appropriately here. The value of iceMask will give some indication of how bad the interpolation ramp is in the thickness field, while you can't know that from looking at the interpolated thickness field alone.

matthewhoffman: Ok, that makes sense, thanks.

PR #469, opened 2022-07-28
Fix bug when looping over variables

trhille:
When var_name was `"beta"`, `extrapolation` was being reset to `None` instead of `"min"` because of the inadvertent `"if"` instead of `"elif"`. Testing confirms this commit fixes the issue and properly sets the `extrapolation` method for `beta` to `"min"`.


matthewhoffman (commented):
Good catch, @trhille .  I guess since we don't use beta much anymore this wasn't on the radar.

PR #470, opened 2022-08-09
Add additional variables when create LI mesh

matthewhoffman:
This merge adds the following variables whenever a new mesh is
created: stiffnessFactor, groundedVonMisesThresholdStress,
floatingVonMisesThresholdStress, iceMask
It also remove the --eigencalving option so that eigencalvingParameter
is always included.  The idea is to avoid a proliferation of options and
err on the side of including all the fields that might be used more
than infrequently.  If a few 2d fields are not used in a mesh, they do
not add a significant amount to the file size.


trhille (approved):
Approved by inspection

PR #471, opened 2022-08-23
Add analysis script for plotting regional stats

matthewhoffman:
This script is modified from the global stats script.  It plots three
different figures, each with subpanels for each region.  The first is
overall mass change statistics.  The second is the grounded mass
balance, and the third is the floating mass balance.  It is possible to
plot up to four runs, each with a different linestyle, but because of
the number of lines on each plot, plotting multiple runs is not pretty.


landice/output_processing_li/plot_regional_stats.py
```diff
@@ -0,0 +1,291 @@
+#!/usr/bin/env python
```
trhille, line 1: Note that the name of this script follows a difference scheme from plot_globalStats.py. For consistency it's probably better to call it plot_regionalStats.py.

landice/output_processing_li/plot_regional_stats.py
```diff
@@ -0,0 +1,291 @@
...
+
+if options.fileRegionNames:
+   fn = Dataset(options.fileRegionNames, 'r')
+   rNames = fn.variables['regionNames'][:]
+
+# Get nRegions from first file
+f = Dataset(options.file1inName, 'r')
+nRegions = len(f.dimensions['nRegions'])
+
+nrow=4
```
trhille, line 51: Can we either have the script calculate the necessary number of rows and columns or have these as user input options?

landice/output_processing_li/plot_regional_stats.py
```diff
@@ -0,0 +1,291 @@
...
+   plt.ylabel('volume change ({})'.format(massUnit))
+   #plt.xticks(np.arange(22)*xtickSpacing)
+   plt.grid()
+   axs1.flatten()[reg].set_title(rNames[reg,:].tobytes().decode('utf-8').strip())
+   if reg == 0:
+      axX = axs1.flatten()[reg]
+   else:
+      axs1.flatten()[reg].sharex(axX)
+
+# Set up Figure 2: grd MB
```
trhille, line 71: can you change "grd MB" to "grounded mass balance"

landice/output_processing_li/plot_regional_stats.py
```diff
@@ -0,0 +1,291 @@
...
+   plt.ylabel('volume change ({})'.format(massUnit))
+   #plt.xticks(np.arange(22)*xtickSpacing)
+   plt.grid()
+   axs2.flatten()[reg].set_title(rNames[reg,:].tobytes().decode('utf-8').strip())
+   if reg == 0:
+      axX = axs2.flatten()[reg]
+   else:
+      axs2.flatten()[reg].sharex(axX)
+
+# Set up Figure 3: flt MB
```
trhille, line 86: can you change "flt MB" to "floating mass balance"

landice/output_processing_li/plot_regional_stats.py
```diff
@@ -0,0 +1,291 @@
...
+
+nrow=4
+ncol=4
+if nRegions > nrow*ncol:
+    sys.exit("ERROR: Number of regions exceeds number of plots.  Please adjust nrow, ncol as needed.")
+
+# Set up Figure 1: volume stats overview
+fig1, axs1 = plt.subplots(nrow, ncol, figsize=(13, 11), num=1)
+fig1.suptitle('Mass change summary', fontsize=14)
+for reg in range(nRegions):
```
trhille, line 59: Instead of a separate loop for each figure, what about using nested loops over figures and regions?

trhille: e.g.,:
```
for ax in [axs1, axs2, axs2]:
    for reg in range(nRegions):
```

landice/output_processing_li/plot_regional_stats.py
```diff
@@ -0,0 +1,291 @@
...
+def plotStat(fname, sty):
+    print("Reading and plotting file: {}".format(fname))
+
+    name = fname
+
+    f = Dataset(fname,'r')
+    yr = f.variables['daysSinceStart'][:]/365.0
+    nRegionsLocal = len(f.dimensions['nRegions'])
+    if nRegionsLocal != nRegions:
+        sys.exit(f"ERROR: Number of regions in file {fname} does not match number of regions in first input file!")
```
trhille, line 113: This will be triggered for any input file, not just the first one, right?

landice/output_processing_li/plot_regional_stats.py
```diff
@@ -0,0 +1,291 @@
...
+
+    # Fig 1: summary plot
+    vol = f.variables['regionalIceVolume'][:]
+    if options.units == "m3":
```
trhille, line 117: Can we loop over a few list of variables based on their units instead of having ten separate instances of this if-statement?

landice/output_processing_li/plot_regional_stats.py
```diff
@@ -0,0 +1,291 @@
...
+    plotStat(options.file2inName, sty=':')
+
+if(options.file3inName):
+    plotStat(options.file3inName, sty='--')
+
+if(options.file4inName):
+    plotStat(options.file4inName, sty='-.')
+
+
+axs1.flatten()[-1].legend(loc='best', prop={'size': 6})
```
trhille, line 282: The legend doesn't show up if the last subplot is empty. For example, I'm using 27 regions with nrow = 7 and nrow = 4, so the -1th subplot has nothing in it. Can we put the legend in the first subplot to avoid that?

trhille (changes requested):
@matthewhoffman This is great. I have requested some changes to reduce redundancy and increase readability, but functionally everything is there. The only functional issue is that the legend doesn't show up if the last subplot is empty (i.e., if there is an odd number of regions).

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
```diff
@@ -0,0 +1,197 @@
...
+parameterization that includes slope.  To get this field, run MALI one timestep with
+config_ismip6shelfMelt_use_slope=.true. and shelfBaseSlope in the output stream.  This will
+ensure that the slope is calculated in an identical way to how MALI would do it.
+
+Note that gamma0 is set in the script, as well as the range of deltaT to search over.
+
+The tuned basin-by-basin deltaT values are written to a file called
+'basin_and_coeff_gamma0_DeltaT_quadratic_non_local.nc'. You will want to rename it to
+avoid clobbering it if the script is rerun.  In addition to saving deltaT, the file also
+include the basin info and gamma0 so it can dropped directly into a streams file to run with.
```
trhille, line 14:
```suggestion
includes the basin info and gamma0 so it can dropped directly into a streams file to run with.
```

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
```diff
@@ -0,0 +1,197 @@
...
+def calcMelt(deltaT):
+    mean_TF = 0.0
+    IS_area = 0.0
+    meanSlope = 0.0
+    TFdraft = np.zeros((nCells,))
+    meanTFcell = np.zeros((nCells,))
+
+    for iCell in range(nCells):
+        if floatMask[iCell] == 1:
+           # 1 -  Linear interpolation of the thermal forcing on the ice draft depth :
```
trhille, line 127: no # 2?

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
```diff
@@ -0,0 +1,197 @@
...
+           TFdraft[iCell] = np.interp(lowerSurface[iCell], np.flip(zOcean), np.flip(TFocean[iCell,:])) # flip b/c z is ordered from sfc to deep but interp needs to be increasing
+    meanTF = np.zeros((nRegions,))
+    for r in range(nRegions):
+        ind = np.nonzero(floatMask * regionCellMasks[:,r])[0]
+        meanTF[r] = (TFdraft[ind]*areaCell[ind]).sum() / areaCell[ind].sum()
+        meanTFcell[ind] = meanTF[r]
+
+    melt = gamma0 * coef * slope * (TFdraft + deltaT) * np.absolute(meanTFcell + deltaT) # m/yr
```
trhille, line 135: Are you sure we want `np.absolute(meanTFcell + deltaT)`? Conceivably meanTFcell + deltaT could be negative, in which case we probably don't want it to end up being positive, right? Instead, do we want `np.maximum(meanTFcell + deltaT, 0.)`?

landice/output_processing_li/plot_globalStats.py
```diff
@@ -49,38 +50,56 @@
 
 axVol = fig.add_subplot(nrow, ncol, 1)
 plt.xlabel('Year')
-plt.ylabel('volume ({})'.format(massUnit))
+if options.plotChange:
```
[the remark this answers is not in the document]
trhille: However, while we are changing this, we might as well switch over to using f-strings.

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
```diff
@@ -0,0 +1,197 @@
...
+        totalMelt[r] = (melt[ind]*areaCell[ind]).sum() * rhoi / 1.0e12 #convert to Gt/yr
+
+    return melt, totalMelt, TFdraft
+
+print("Considering dTs", dTs)
+allMelts = np.zeros((len(dTs), nRegions))
+for i, dT in enumerate(dTs):
+    print(f"Calculating with dT={dT}")
+    melt, allMelts[i,:], TFdraft = calcMelt(dT)
+    if False:
```
trhille, line 149: Let's have a `-p` option to determine whether to plot or not.

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+print("** Gathering information.  (Invoke with --help for more details. All arguments are optional)")
+parser = OptionParser(description=__doc__)
+parser.add_option("-g", dest="fileName", help="input filename that includes ice geometry information. If shelfBaseSlope is included in this file, the script will assume the form of the melt parameterization that includes slope", metavar="FILENAME")
+parser.add_option("-n", dest="fileRegionNames", help="region name filename.", metavar="FILENAME")
+parser.add_option("-o", dest="fcgFileName", help="ocean forcing filename", metavar="FILENAME")
+options, args = parser.parse_args()
+
+ISMIP6basinInfo = {
```
trhille, line 54: These are from Rignot et al. (2008) and Rignot et al. (2013)? Needs a comment.

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
```diff
@@ -0,0 +1,197 @@
...
+       plt.scatter(xCell, yCell, s=1, c=melt, cmap='jet')
+       plt.title(f'dT={dT}')
+       plt.colorbar()
+
+
+bestdTCells = np.zeros((nCells,))
+regionCells = np.zeros((nCells,), 'i')
+bestdT = np.zeros((nRegions,))
+for r in range(nRegions):
+    bestdT[r] = np.interp(ISMIP6basinInfo[rNamesOrig[r]]['shelfMelt'][0], allMelts[:,r], dTs)
```
trhille, line 160:
```suggestion
    # Find dT that results in Rignot et al. (2008) melt rate for each basin
    bestdT[r] = np.interp(ISMIP6basinInfo[rNamesOrig[r]]['shelfMelt'][0], allMelts[:,r], dTs)
```

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
```diff
@@ -0,0 +1,197 @@
...
+    print(f"{ISMIP6basinInfo[rNamesOrig[r]]['name']}: {bestdT[r]}")
+    bestdTCells[regionCellMasks[:,r]==1] = bestdT[r]
+    regionCells[regionCellMasks[:,r]==1] = r+1
```
trhille, line 163: Why is this done?

trhille:
@matthewhoffman, I've reviewed the two new pre-processing scripts. I still need to re-review plot_regional_stats.py, but I wanted to look those two over to test out https://github.com/MALI-Dev/E3SM/pull/48

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
matthewhoffman: That's by design.  It allows you to get freeze-on when TF<0.  If you look at the PIGL plots in the Jourdain paper, you can see negative melt rates in some places.

trhille: Yes that makes sense now. I had tunnel vision on that last term.

landice/output_processing_li/plot_regional_stats.py
matthewhoffman: At this point (might not have been the the case when you reviewed this in August...) there are enough differences between figures that I think it's easier to leave as separate chunks of code for each.

landice/output_processing_li/plot_regional_stats.py
matthewhoffman: Yes, it will be triggered for any input file.  It doesn't seem possible to ensure the regions are the same regions, but we can at least err if the number of regions doesn't match.

landice/mesh_tools_li/adjust_iceshelf_geometry.py
```diff
@@ -0,0 +1,128 @@
...
+oMgnMask = np.zeros((nCells,),'i')
+for c in range(nCells):
+    if floatMask[c] == 1:
+        for n in range(nEOnC[c]):
+            if thickness[cOnC[c,n]-1] == 0.0:
+                oMgnMask[c] = True
+                break
+
+oMgnMaskOld = oMgnMask.copy()
+for c in range(nCells):
```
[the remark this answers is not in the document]
matthewhoffman: I've added a comment to the code explaining this step.

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
matthewhoffman: In general, I don't think we will want to plot a melt map for every dT considered, but it can be useful for debugging, etc.  I added a comment to that effect.

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
matthewhoffman: the ISMIP6 region mask is a single integer field, while regionCellMasks is a separate 0/1 mask for each region, so this converts the latter to the former.  I added a comment to the code about it.

matthewhoffman:
@trhille , now that things have settled down with our ISMIP6 workflow, I finally cleaned up this PR.  I deleted a few scripts that did not end up being relevant, and I responded to review comments and did general cleanup with the others.  I responded directly to comments above for which I had a response.  For those without a response, I simply made the changes.  

I realize there is a lot to review in this PR; if you prefer this be broken into a couple PRs for different scripts, I can do that.  Also, there is no real rush to get this re-reviewed and merged, but it would be nice to check this off and have it on master before too much time goes on.  I'll leave it up to you regarding the level of detail you think further review warrants.

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
```diff
@@ -0,0 +1,243 @@
...
+rNames = [None]*nRegions
+for reg in range(nRegions):
+    if rNamesOrig[reg] in ISMIP6basinInfo:
+        rNames[reg] = ISMIP6basinInfo[rNamesOrig[reg]]['name']
+    else:
+        rNames[reg] = rNamesOrig[reg]
+
+ff = Dataset(options.fcgFileName, 'r')
+zOcean = ff.variables['ismip6shelfMelt_zOcean'][:]
+TFocean = ff.variables['ismip6shelfMelt_3dThermalForcing'][0,:,:]
```
trhille, line 104: We might want to either make a note in the header docstring that this operation is on the zeroth time level of the forcing file, or add the time level as an optional user input that defaults to 0. I can see wanting to use this script on the year 2015, for example, with a forcing file that starts in 1950.

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
```diff
@@ -0,0 +1,243 @@
...
+nCells = len(f.dimensions['nCells'])
+areaCell = f.variables['areaCell'][:]
+xCell = f.variables['xCell'][:]
+yCell = f.variables['yCell'][:]
+if options.use_slope:
+    print('USING SLOPE')
+    slope = f.variables['shelfBaseSlope'][1,:]
+    ind = np.nonzero(floatMask==1)[0]
+    meanSlope = (slope[ind] * areaCell[ind]).sum() / areaCell[ind].sum()
+    meanSlopeFactor = ((1.0 + 900.0 * slope[ind]) * areaCell[ind]).sum() / areaCell[ind].sum()
```
trhille, line 127: Add comments describing where `meanSlopeFactor` and `meanSlopeFactor2` calculations come from.

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
```diff
@@ -0,0 +1,243 @@
...
+    meanSlopeFactor2 = (np.exp2(np.log10(np.maximum(1.0e-4,slope[ind]))+4) * areaCell[ind]).sum() / areaCell[ind].sum()
+    print(f"Mean slope = {meanSlope}, mean slope factor = {meanSlopeFactor}")
+    print(f"Mean slope = {meanSlope}, mean slope factor2 = {meanSlopeFactor2}")
+else:
+    slope = np.zeros((nCells,))
+
+def calcMelt(deltaT):
+    mean_TF = 0.0
```
trhille, line 135: This definition is never used.
```suggestion

```

landice/output_processing_li/plot_calving_calibration.py
```diff
@@ -0,0 +1,244 @@
...
+   plt.xlabel('Year')
+   plt.ylabel('VAF change ({})'.format(massUnit))
+   plt.grid()
+   axs3.flatten()[reg].set_title(rNames[reg])
+   if reg == 0:
+      axX = axs3.flatten()[reg]
+   else:
+      axs3.flatten()[reg].sharex(axX)
+
+# Set up Figure 2: calving
```
trhille, line 146:
```suggestion
# Set up Figure 4: calving
```

landice/output_processing_li/plot_regionalStats.py
```diff
@@ -0,0 +1,431 @@
...
+# Antarctic data from:
+# Rignot, E., Bamber, J., van den Broeke, M. et al. Recent Antarctic ice mass loss from radar interferometry
+# and regional climate modelling. Nature Geosci 1, 106-110 (2008). https://doi.org/10.1038/ngeo102
+# Table 1: Mass balance of Antarctica in gigatonnes (10^12 kg) per year by sector for the year 2000
+# https://www.nature.com/articles/ngeo102/tables/1
+# and
+# Rignot, E., S. Jacobs, J. Mouginot, and B. Scheuchl. 2013. Ice-Shelf Melting Around Antarctica. Science 341 (6143): 266-70. https://doi.org/10.1126/science.1235798.
+# Note: May want to switch to input+, net+
+# Note: Some ISMIP6 basins combine multiple Rignot basins.  May want to separate if we update our regions.
+ISMIP6basinInfo = {
```
trhille, line 79: We should have a note in the header docstring that explains that currently the script is only useful for whole Antarctic runs. Maybe in a future PR we can add an optional input variable that points to a file that contains observational data, so this can be used for any arbitrary configuration, with or without observations plotted.

landice/output_processing_li/plot_regionalStats.py
```diff
@@ -0,0 +1,431 @@
...
+        axs3.flatten()[r].plot(yr, cumFltSum2[:,r], label=lbl, linestyle=':', color='hotpink', linewidth=0.75)
+
+
+    # Fig 4: area change  ---------------
+
+    areaTot = f.variables['regionalIceArea'][:]/1000.0**2
+    areaGrd = f.variables['regionalGroundedIceArea'][:]/1000.0**2
+    areaFlt = f.variables['regionalFloatingIceArea'][:]/1000.0**2
+    for r in range(nRegions):
+        axs4.flatten()[r].plot(yr, areaTot[:,r] - areaTot[0,r], label=("total area" if addToLegend else '_nolegend_'), linestyle=sty, color='k')
```
trhille, line 376: Definition of `lbl` in figures above is better practice than the repetition of `label=("total area" if addToLegend else '_nolegend_')` here.

trhille (changes requested):
@matthewhoffman, this is great! I have a few fairly minor comments to address before we merge this. Thanks for all your hard work on this.

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
```diff
@@ -0,0 +1,243 @@
...
+
+# Antarctic data from:
+# Rignot, E., Bamber, J., van den Broeke, M. et al. Recent Antarctic ice mass loss from radar interferometry
+# and regional climate modelling. Nature Geosci 1, 106-110 (2008). https://doi.org/10.1038/ngeo102
+# Table 1: Mass balance of Antarctica in gigatonnes (10^12 kg) per year by sector for the year 2000
+# https://www.nature.com/articles/ngeo102/tables/1
+# and
+# Rignot, E., S. Jacobs, J. Mouginot, and B. Scheuchl. 2013. Ice-Shelf Melting Around Antarctica. Science 341 (6143): 266-70. https://doi.org/10.1126/science.1235798.
+# Note: Only basin names and shelfMelt fields used in this script.
+ISMIP6basinInfo = {
```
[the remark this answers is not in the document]
trhille: This would also be useful for generalizability.

landice/mesh_tools_li/bulldoze_missing_troughs.py
```diff
@@ -0,0 +1,151 @@
...
+        fracWidth = min(fracWidth, 0.95)
+        print(f'found {len(ind2)} points along this orthogonal; width={width}; fracWidth={fracWidth}, mag={mag}')
+        z = mag * (-1.0 * fracWidth**2 + 1.0)
+        WCTnew[c] = max(z, WCT[c])  # don't make WCT thinner than the original
+        print(f'old z={WCT[c]}, new z={z}')
+
+    return WCTnew
+
+
+# Define maximum and minimum water column thickness along tough centerline
```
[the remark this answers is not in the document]
matthewhoffman: I like leaving them here, because they go along with the point definitions to be modified.  Also note that the final trough is hardcoded to use different values.  I added a note to the docstring directing a user to that part of the code.

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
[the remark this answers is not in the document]
matthewhoffman: Yeah, it would be nice to put it some place central so we aren't repeating it, but I punted on dealing with that for now.  If all this was in the conda package, we could use it, but for the scripts as written, the most robust thing seemed to be to include it everywhere.

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
matthewhoffman: I added a note in the docstring and the argument description for the TF file.  I imagine in general we'll want to apply this to a climatology rather than a specific year, but we can add an option if more flexibility is desired.

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
matthewhoffman: I deleted all the slope stuff, because that branch never survived in MALI.  The slope version of this script will be in the history of this branch if we ever need it again.

landice/mesh_tools_li/tune_ismip6_melt_deltat.py
matthewhoffman: removed

landice/output_processing_li/plot_regionalStats.py
matthewhoffman: note added

landice/output_processing_li/plot_regionalStats.py
matthewhoffman: In this case there'd be 3 different `lbl` variables defined before the for loop and used inside it, so I left this easier to read version.

matthewhoffman:
@trhille , I pushed another cleanup commit addressing your latest round of review.  I think this should be ready to go now.

trhille (approved):
Thanks, @matthewhoffman, for this great set of tools!

PR #472, opened 2022-09-01
Change keepMask for extrapolating stiffnessFactor

matthewhoffman:
Before the keepMask for stiffnessFactor was thickness>0.  However, this
excluded the extended Albany cell.  This commit changes the logic to
make the keepMask where the optimization returned values of
stiffnessFactor>0 (i.e. the whole Albany mesh).  There still is
extrapolation into the MPAS mesh beyond the Albany mesh.  That
extrapolation is adjusted to disable smoothing.  There is no reason to
think smoothing is of value there, and it takes longer to do it.

Also updated the help.


trhille (approved):
Approved by inspection.

PR #474, opened 2022-09-07
Replace `$` with `$$` when writing config options

### LINKED ISSUE #473: `$$` getting written as `$` in output config file
It is not possible to use the output config file (the one in `logs` or the complete config file in `html`) as an input config file because certain `$$` in strings get written out instead as `$`.  Then, on reading these back in, they get interpreted as references to other config options.

We need to figure out a way to output `$` as `$$`.  This may need to happen in MPAS-Tools, rather than here.

Thanks @akturner for reporting this issue.

xylar:
This is necessary because we always use extended interpolation when we read config files, and a single `$` is interpreted as a reference to another config option.

closes #473

### DIFF
--- a/conda_package/mpas_tools/config.py
+++ b/conda_package/mpas_tools/config.py
@@ -364,7 +364,7 @@ class MpasConfigParser:
                 if include_sources:
                     source = self.sources[(section, option)]
                     fp.write(f'# source: {source}\n')
-                value = str(value).replace('\n', '\n\t')
+                value = str(value).replace('\n', '\n\t').replace('$', '$$')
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I tested MPAS-Analysis with this fix.  Without the fix, the output config files contain values with `$` and with the fix they instead have `$$` as expected.  Further, I made sure that I could run `mpas_analysis` with the config file written out by a previous run (altering the output path to avoid collisions).  Neither analysis run has finished yet but both are running smoothly so far.

I will also test compass but I do not expect any difficulty there, since compass does not include config options with `$$`

PR #475, opened 2022-09-19
Add sea-level equiv axis to VAF plot

trhille:
Add a sea-level equivalent axis in mm to VAF subplot in plot_globalStats.py.
Also remove some redundancy in unit conversions.

image

### DIFF
--- a/landice/output_processing_li/plot_globalStats.py
+++ b/landice/output_processing_li/plot_globalStats.py
@@ -13,7 +13,7 @@ from optparse import OptionParser
 import matplotlib.pyplot as plt
 
 rhoi = 910.0
-
+rhosw = 1028.
 
 print("** Gathering information.  (Invoke with --help for more details. All arguments are optional)")
 parser = OptionParser(description=__doc__)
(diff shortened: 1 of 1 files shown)


landice/output_processing_li/plot_globalStats.py
```diff
@@ -97,6 +100,15 @@
 plt.grid()
 
 
+def VAF2seaLevel(vol):
+    return vol * scaleVol / 3.62e14 * rhoi / rhosw * 1000.
```
matthewhoffman, line 104: What do you think about making this always relative to the initial sea level equivalent? i.e.
```suggestion
    return (vol - vol[0]) * scaleVol / 3.62e14 * rhoi / rhosw * 1000.
```
(and similar for the inverse)

trhille: @matthewhoffman, this would only work if all the runs plotted are starting from the same VAF, which won't always be the case. So we can either have both axes show totals or change from initial, but we can't have the VAF axis show total and the SLR axis show change from initial if we want to maintain generality. I'd be fine with either way. Let me know which one you'd prefer.

trhille: @matthewhoffman, are you happy with this, or would you like me to make both axes show change from the initial time?

trhille:
@matthewhoffman, I think the new `plotChange` option in https://github.com/MPAS-Dev/MPAS-Tools/pull/471 will give you the functionality you asked for in this PR.

matthewhoffman (approved):
@trhille , yes, in combination with #471 it is possible to plot SLR as change from initial.  I rebased and force-pushed to resolve conflicts arising from the recently merged #471.

PR #483, opened 2022-12-12
Constrain shapely because of big changes in 2.0

xylar:
Here is a description of what we will need to change to be able to work with 2.0:
https://shapely.readthedocs.io/en/stable/migration.html


xylar:
Shapely 2.0 has just been released (https://github.com/conda-forge/shapely-feedstock/pull/120) and it will likely break MPAS-Tools.

PR #484, opened 2022-12-12
Reorganize sea-ice grid tools into python package

xylar:
This merge moves many functions from `seaice_grid_tools` into a `seaice` python package.  This is intended as an intermediate step (easier to review) before moving the `seaice` package within the `mpas_tools` package.

A `CMake` build capability has been added in anticipation of building `fix_regrid_output.exe` for the conda package.

Some fairly minimal clean-up has been done as part of the proposed reorganization.

Many python scripts in `seaice_grid_tools` are now just stubs that call functions within the `seaice` package.


akturner (approved):

xylar:
Thanks @akturner!

PR #489, opened 2023-01-04
Fix importlib.resources imports for python 3.8

xylar:
This merge also updates:
* Require `importlib_resources` for python 3.8
* Required `geometric_features` (to fix the same issue)


xylar:
## Testing

Successfully built the conda package (which runs `pytest`) with python 3.8.

PR #492, opened 2023-02-12
Fix reading of SciVisColor colormaps

xylar:
The old approach (with `pkg_resources`) is considered to be a bad practice and was giving me trouble in `polaris`, which has `mpas_tools` as a dependency.

This merge also updates to v0.18.0 for a release


xylar:
## Testing

With the new import approach using `importlib.resources`, things worked as expected and I get no complaints.

PR #499, opened 2023-03-10
Allow external mapping file for sea-ice partitioning

### LINKED ISSUE #497: sea ice partitioning tool needs to take an MPI exec and core count
The sea-ice partitioning tool currently calls `ESMF_RegridWeightGen` but it doesn't have a way to call it with a parallel executable or with a number of cores.  On some machines, even a 1-core call has to be made with a parallel executable.  And it would be nice to use multiple cores if available.

For task parallelism, it may also be necessary to break those calls out into their own separate steps, but that can be left for future work.

xylar:
This allows us to generat the mapping file externally if needed (e.g. if calling ESMF requires a parallel executable like `srun` and if we want to use MPI parallelism).

closes #497

### DIFF
--- a/conda_package/mpas_tools/seaice/partition.py
+++ b/conda_package/mpas_tools/seaice/partition.py
@@ -189,16 +189,26 @@ def prepare_partitions():
     """
     # parsing input
     parser = argparse.ArgumentParser(
-        description='Perform preparatory work for making seaice partitions.')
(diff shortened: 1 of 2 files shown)


xylar:
## Testing

I was able to use this branch along with https://github.com/MPAS-Dev/compass/pull/559 to run the QU240 `files_for_e3sm` test case on both Chrysalis and Anvil.

PR #500, opened 2023-03-12
Update docs for seaice partitioning

xylar:


xylar:
@maltrud and @akturner, letting you both know I have updated the docs as promised.

PR #502, opened 2023-03-25
Add an optional timeout to check_call with logging

xylar:
This is useful for killing a process that is hanging, for example in debugging.


xylar:
## Testing

I tested this in `compass`, where I set `timeout=120` to debug a test case that is hanging.  As expected, the timeout was reached and the test case was killed with an error message.

PR #505, opened 2023-04-21
Add plotting tool for MPAS-Ocean transects 

xylar:
They can be plotted either with `mpas_tools.ocean.viz.plot_ocean_transects()` or a command-line tool `plot_ocean_transects`.


xylar:
@dengwirda, would you be willing to give this a try?  Testing would involve checking out this branch, activating your conda base environment and then running:
```
mamba create -y -n mpas_dev mpas_tools cmocean
mamba activate mpas_dev
python -m pip install -e conda_package
```

Then, you would need to create set of transects.  I like to use https://geojson.io/, and then add in the 4 required properties by hand.  Here's what I used:
```json
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {
          "name": "Weddell to FRIS",
          "object": "transect",
          "component": "ocean",
          "author": "Xylar Asay-Davis"
      },
      "geometry": {
        "coordinates": [
          [
            -24.471951656472072,
            -56.48794865789691
          ],
          [
            -31.660601080601253,
            -68.39124416042628
          ],
          [
            -98.08418099854232,
            -85.05112877980659
          ]
        ],
        "type": "LineString"
      }
    },
    {
      "type": "Feature",
      "properties": {
          "name": "South Atlantic",
          "object": "transect",
          "component": "ocean",
          "author": "Xylar Asay-Davis"
      },
      "geometry": {
        "coordinates": [
          [
            -44.34409086044417,
            -27.918704425783957
          ],
          [
            -15.762534619285049,
            -31.82980253026402
          ],
          [
            13.163988726046568,
            -21.932286503024528
          ]
        ],
        "type": "LineString"
      }
    }
  ]
}
```

I've only tested on a tiny file (QU240 mesh) so far, but I'm optimistic that it's still a good test.  I'll try the EC30to60 and ECwISC30to60 shortly.

One more note: I decided to plot the MPAS cells as having constant properties for now, rather than linearly interpolating between cell values.  That's relatively easy to change (or provide a command-line flag to control).  Let me know what you think.

xylar:
Weddell_to_FRIS_temperature
Weddell_to_FRIS_salinity
Weddell_to_FRIS_layerThickness
South_Atlantic_temperature
South_Atlantic_salinity
South_Atlantic_layerThickness

xylar:
I need to add the red and green bars to the left and right y axes so it's clear which is which.

milenaveneziani:
Nice @xylar!
Maybe we could add interpolation for `normalVelocity`?

xylar:
@milenaveneziani, I really don't know how to do that.  I've never plotted quantities on edges with this approach and I don't think it would be trivial to support.  You could plot the reconstructed zonal and meridional velocities or we could figure out a way to project those to be normal to the transect but this approach doesn't follow model edges, it really slices through cells in exactly the path the transect takes.  So it's hard to get normal velocity.  (Plus, normal velocity is an absolute mess!  I know you're used to looking at it, but still...)

milenaveneziani:
ok, thanks for clarifying that @xylar. Maybe the zonal and meridional vel components is the way to go, at least for visualization purposes. When we compute stuff, such as transports, then we definitely want to use `normalVelocity` but for plots I agree we could be less rigorous.

May I add a suggestion, if possible at all (and easy)? would it be straightforward to add an option to flip the transects? This is just because sometimes we are used to look at transects in the opposite direction wrt what the geojson file has in it. Again, not a priority though.

xylar:
@milenaveneziani, given that one `geojson` file has many transects, how would you want to indicate which to flip?  Or woudl you want to flip them all?

xylar:
@milenaveneziani, I went ahead and added a `--flip` flag.  I figure you could always break your plots into 2 geojson files, one to flip and one not to if you have to.

milenaveneziani:
hmm, didn't think of that. seems cumbersome. I think the main 'problem' is that we are used to see zonal transects as going west to east, but that is usually opposite to the direction of geojson files.. (whereas south to north for meridional transects works fine). Maybe better to let it be.

xylar:
@milenaveneziani, one more thing.  A big part of the purpose of this was that you could always make your own copy of `conda_package/mpas_tools/ocean/viz/transects.py` and either leave it as a script of make it a Jupyther notebook.  Then, you can go into `_plot_transect()` and change whatever you need to to customize things (e.g. figure size, dpi, colorbar...).

xylar:
@dengwirda, I need to make a release so I'm going to merge this.  Sorry I didn't wait for your review but I hope my testing was sufficient.

@milenaveneziani, thanks for your feedback!

PR #509, opened 2023-05-02
Fixes to ISMIP6 processing for scalar fields

matthewhoffman:
This PR introduces additional fixes to the ISMIP6 post-processing tool for MALI.  It corrects the handling of time levels for scalar state and flux fields, and it adds a missing scalar field that previously was not possible to calculate but now we can.


matthewhoffman:
@hollyhan , here are the changes for the scalar files.  We can add your helper script for converting a restart file to a state file when it's ready.

PR #510, opened 2023-05-18
Fix culling in VTK extractor with multiple times

xylar:
If here are multiple time slices per file and you try to cull the files to a region, we want to only cull the unique files. Previously, the tool was culling the same file repeatedly, once for each time index.


xylar:
@alicebarthel, when you can, could you review this?  I'd like to include it in the next MPAS-Tools release.

xylar:
@alicebarthel, given that you're away, I'm going to merge this and hope that my testing is sufficient.

PR #517, opened 2023-08-04
Fix widen transect function

xylar:
The mask should be of int type, not float


xylar:
Without this fix, I'm seeing
```
double transectCellMasks(nCells, nTransects)
```
but with the fix, I see:
```
int transectCellMasks(nCells, nTransects)
```
as expected.

The mask being of type double was causing segfaults in conjunction with #514

xylar:
Tested along with #514 in compass, and ran successfully.

PR #515, opened 2023-07-06
Add bandwidth-minimising mesh sorting routine

dengwirda:
This PR adds a mesh sorting routine `mpas_tools.mesh.creation.sort_mesh` that can be used to reindex an MPAS mesh file to bring the cells, edges and vertices into a more cache-local sequence. This undoes the somewhat 'random' ordering introduced by the mesh conversion tools, and may be desirable re: efficiency for downstream dycore + coupler applications.

The reordering is implemented via a reverse cuthill-mckee approach --- minimising the bandwidth of the cell-to-cell adjacency graph and then permuting the edges and vertices to match. Sorting both culled and unculled meshes is supported, as well as meshes containing additional variables, which are reindexed according to their underlying `nCells`, `nEdges` or `nVertices` dimensions.

The mesh sorter should be callable as a cmd-line tool:
```
python sort_mesh.py \
--mesh-file=path+name-to-mpas-file-for-sort \
--sort-file=path+name-to-sorted-output-file
```
and/or inline:
```
from mpas_tools.mesh.creation.sort_mesh import sort_mesh
mesh = xarray.open_dataset("mpas-file-name")
sort_mesh(mesh)
```
Tested on `pm-cpu` using a QU30km MPAS-O standalone config., which successfully spun-up to 90 days after a sort of the culled mesh. Sorting improves the locality of the various remapping arrays (`verticesOnCell`, `edgesOnVertex`, etc, etc) with all showing smooth distributions.

`edgesOnCell=unsorted`:
eoc_unsorted

`edgesOnCell=issorted`:
eoc_issorted

PR #518, opened 2023-08-07
Fix MOC edge sign and mask

xylar:
Testing of #515 has exposed some issues with the code for computing MOC southern transects.  

This fixes an attempt to mask out a dummy edge, but there is, in fact, no such dummy edge in the data written to files (just in the MPAS code).

This merge also fixes a related bug where the edge sign of edge -1 was being checked, whereas we want to ignore any `edgesOnVertex` that are -1.


xylar:
## Testing

With this fix, `files_for_e3sm/diagnostic_masks` completes successfully for QU240, QUwISC240, EC30to60 and ECwISC30to60.

PR #521, opened 2023-08-17
Convert masks to int type in `cull()` function

xylar:
The new version of the cell culler requires masks in `int` (rather than `int64` or `float`) format.

This is needed because output in NETCDF4 format is producing int64, rather than int32, arrays but the cell culller is then reading a full int64 data array into memory allocated for an int32 array, thus overwriting a lot of memory allocated for other arrays with garbage.


xylar:
## Testing

I was able to cull the mesh in the QUwISC240 mesh test case in compass with this branch, whereas it was crashing without these changes.

PR #526, opened 2023-09-21
Switch from use of 'nc-config' to 'nf-config' in top-level Makefile for grid_rotate

mgduda:
This PR switches to the use of `nf-config` in the top-level `Makefile` for the grid_rotate tool.

Beginning with NetCDF-C 4.9.2, the `nc-config` program no longer provides information from `nf-config`; see
Unidata/netcdf-c@6f55c852

Now, `nf-config`, rather than `nc-config` is used to determine the Fortran compiler, include paths, and library paths to use.

### DIFF
--- a/mesh_tools/grid_rotate/Makefile
+++ b/mesh_tools/grid_rotate/Makefile
@@ -1,7 +1,7 @@
-FC = $(shell nc-config --fc)
+FC = $(shell nf-config --fc)
 FFLAGS = -O3
-FCINCLUDES = $(shell nc-config --fflags)
-FCLIBS = $(shell nc-config --flibs)
+FCINCLUDES = $(shell nf-config --fflags)
+FCLIBS = $(shell nf-config --flibs)
 
 all: grid_rotate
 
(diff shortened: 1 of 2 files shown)


xylar:
@mgduda, it seems like maybe the README should be fixed, too?
https://github.com/MPAS-Dev/MPAS-Tools/blob/009168c7bcacd853c1681c505d5a91ec65ecea07/mesh_tools/grid_rotate/README#L30-L36

dimomatt:
Looks good once the README is updated- does the revision history also need an entry?

mgduda:
> @mgduda, it seems like maybe the README should be fixed, too?
> 
> https://github.com/MPAS-Dev/MPAS-Tools/blob/009168c7bcacd853c1681c505d5a91ec65ecea07/mesh_tools/grid_rotate/README#L30-L36

Good catch! I've just force-pushed an update the `README`.

xylar (approved):

mgduda:
> Looks good once the README is updated- does the revision history also need an entry?

I've updated the revision history in the `README` as well. We could probably consider just removing that section of the `README` at some point, as anyone interested the history of the code could just browse the git log.

dimomatt (approved):

PR #527, opened 2023-09-27
Add new features to config parser

xylar:
This merge adds 2 new features to the `MpasConfigParser`:
* Config files can be written out in "raw" format, which preserves the config options that use extended interpolation, rather than substituting them.  This is useful in situations where you want to write out a config file that can still take advantage of extended interpolation when it is read back in.  Writing out in raw format is now the default.
* Two new methods, `append()` and `prepend()` allow you to add a copy of another set of config options either before or after a given set of options.  This is useful for combining a set of specialize options with a set of common or default options, particularly if this procedure needs to happen repeatedly (e.g. for separate tasks, each with its own specialized options).

### DIFF
--- a/conda_package/docs/api.rst
+++ b/conda_package/docs/api.rst
@@ -116,6 +116,8 @@ Config
    MpasConfigParser.set
    MpasConfigParser.write
    MpasConfigParser.copy
+   MpasConfigParser.append
+   MpasConfigParser.prepend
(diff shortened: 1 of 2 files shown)


xylar:
This work has been tested along with this Polaris branch:
https://github.com/xylar/polaris/tree/add-shared-configs
This testing exercised the new raw writing feature and the `prepend()` method.

PR #529, opened 2023-09-29
Fix append and prepend for the `MpasConfigParser`

xylar:
We need to mark the combined config options as invalid after appending or prepending.


xylar:
This fixes problems I was seeing in https://github.com/E3SM-Project/polaris/pull/125 where config options were not being updated after a `prepend()` call.

PR #531, opened 2023-10-15
Move vector reconstruction into the conda package

xylar:
This includes a lot of stylistic clean up and adding an entry point `vector_reconstruct`.  The old `operators/vector_reconstructions.py` is left as a stub.


xylar:
## Testing

I was able to use this tool along with https://github.com/MPAS-Dev/compass/pull/717 to create reconstruction coefficients and then use them to reconstruct the Cartesian, zonal and meridional components of the velocity on a QU240 mesh after a short dynamic adjustment.

PR #534, opened 2023-10-28
Fixes to transect extraction and plotting

xylar:
This merge fixes several issues with transects:
* Sometimes, when a segment of a transect ends exactly at an intersection with a cell edge, there are redundant intersection points in the list.  This merge removes those redundant points.
* This merge also fixes some index arrays that were previously type float but should be int.
* Finally, this merge fixes masking of the SSH to ignore cells where `maxLevelCell==0`


xylar:
## Testing

I used this branch along with https://github.com/xylar/polaris/tree/add-isomip-plus-init to produce the following transect plots.  Without these fixes, the plot could not be produced.  Without the SSH fix, transects include a jump to SSH = 0 in the region of grounded ice.

temperatureSection
temperatureSection

PR #540, opened 2023-11-22
Add script to plot transects from MALI output

trhille:
This merge adds a script to plot ice surfaces, bed topography, speed and optionally temperature along a transect. The transect can either be defined using `-x` and `-y` options from the command line, or using `-c` to point to a csv file containing x and y coordinates in the first and second columns, respectively, with no header row. Use `--temperature` if you wish to plot `temperature` along the transect. Note that this uses the 3D `temperature` field and ignores `surfaceAirTemperature` and `basalTemperature`, but these could be added in the future. You can plot more than one time index, specified using the `-t` option; however, plotting temperature at multiple times is not supported. The script increases the number of specified sample points to match the highest resolution in the mesh, based on `dcEdge`, and uses linear interpolation to produce high-resolution transects from a minimal number of specified points (minimum is 2). Temperature is interpolated in x and y, but not in the vertical dimension.

### DIFF
--- a/landice/output_processing_li/plot_transect.py
+++ b/landice/output_processing_li/plot_transect.py
@@ -121,15 +121,15 @@ else:
 # increase sampling to match highest mesh resolution
 total_distance, = np.cumsum( np.sqrt( np.diff(x)**2. + np.diff(y)**2. ) )
 n_samples = int(round(total_distance / np.min(dataset.variables["dcEdge"][:])))
-xArray = np.interp(np.linspace(0, len(x)-1, n_samples),
+x_interp = np.interp(np.linspace(0, len(x)-1, n_samples),
                      np.linspace(0, len(x)-1, len(x)), x)
-yArray = np.interp(np.linspace(0, len(y)-1, n_samples),
+y_interp = np.interp(np.linspace(0, len(y)-1, n_samples),
                      np.linspace(0, len(y)-1, len(y)), y)
 
-d_distance = np.zeros(len(xArray))
-for ii in np.arange(1, len(xArray)):
-    d_distance[ii] = np.sqrt( (xArray[ii] - xArray[ii-1])**2 +
-                              (yArray[ii] - yArray[ii-1])**2 )
+d_distance = np.zeros(len(x_interp))
+for ii in np.arange(1, len(x_interp)):
+    d_distance[ii] = np.sqrt( (x_interp[ii] - x_interp[ii-1])**2 +
+                              (y_interp[ii] - y_interp[ii-1])**2 )
 
 distance = np.cumsum(d_distance) / 1000.  # in km for plotting
 
@@ -143,12 +143,12 @@ timeColors = cm.plasma(np.linspace(0,1,len(times)))
 plt.rcParams.update({'font.size': 16})
 
 bed_interpolant = LinearNDInterpolator(np.vstack((xCell, yCell)).T, bedTopo)
-bed_transect = bed_interpolant(np.vstack((xArray, yArray)).T)
+bed_transect = bed_interpolant(np.vstack((x_interp, y_interp)).T)
 
 for i, time in enumerate(times):
     thk_interpolant = LinearNDInterpolator(
                         np.vstack((xCell, yCell)).T, thk[i,:])
-    thk_transect = thk_interpolant(np.vstack((xArray, yArray)).T)
+    thk_transect = thk_interpolant(np.vstack((x_interp, y_interp)).T)
     lower_surf = np.maximum( -910. / 1028. * thk_transect, bed_transect)
     lower_surf_nan = lower_surf.copy()  # for plotting
     lower_surf_nan[thk_transect==0.] = np.nan
@@ -161,7 +161,7 @@ for i, time in enumerate(times):
     if plot_speed:
         speed_interpolant = LinearNDInterpolator(
                                 np.vstack((xCell, yCell)).T, speed[i,:])
-        speed_transect = speed_interpolant(np.vstack((xArray, yArray)).T)
+        speed_transect = speed_interpolant(np.vstack((x_interp, y_interp)).T)
         speed_transect[thk_transect == 0.] = np.nan
         speedAx.plot(distance, speed_transect, color=timeColors[i])
 
@@ -177,14 +177,14 @@ for i, time in enumerate(times):
                                                      layer_thk[ii,0:-1]) / 2.
             layer_interfaces[ii,:] = upper_surf[ii] - layer_thk[ii,:]
 
-        temp_transect = np.zeros((len(xArray), nVertLevels))
+        temp_transect = np.zeros((len(x_interp), nVertLevels))
         for lev in range(nVertLevels):
             print(f'Interpolating temperature for level {lev}')
             temp_interpolant = LinearNDInterpolator(
                                     np.vstack((xCell, yCell)).T,
                                     temperature[i,:,lev])
             temp_transect[:, lev] = temp_interpolant(
-                                        np.vstack((xArray, yArray)).T)
+                                        np.vstack((x_interp, y_interp)).T)
 
 thickAx.plot(distance, bed_transect, color='black')
 
--- a/landice/output_processing_li/plot_transect.py
+++ b/landice/output_processing_li/plot_transect.py
@@ -112,12 +112,14 @@ if options.coords_file is not None:
     if [options.x_coords, options.y_coords] != [None, None]:
         print('-c and -x/-y options were both provided. Reading from ',
               f'{options.coords_file} and ignoring -x and -y settings.')
+    x = np.asarray(x)
+    y = np.asarray(y)
 else:
-    x = [float(i) for i in options.x_coords.split(',')]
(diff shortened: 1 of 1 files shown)


trhille:
Testing
-------
A transect down the middle of Thwaites Glacier from an ISMIP6 run in 2175:
`python plot_transect.py -d /pscratch/sd/t/trhille/ISMIP6/archive/expAE03/rst.2175-01-01.nc -x -1589311.171,-1208886.353 -y -473916.7182,-357411.6176  --temperature
image

Transects down the middle of Thwaites Glacier from an ISMIP6 run in 2175 and 2179:
`python plot_transect.py -d /pscratch/sd/t/trhille/ISMIP6/archive/expAE03/output/output_state_2175.nc -t 0,-1 -x -1589311.171,-1208886.353 -y -473916.7182,-357411.6176 `
image

landice/output_processing_li/plot_transect.py
```diff
@@ -0,0 +1,214 @@
...
+"""
+Plot bed and surface topography, surface speed, and
+(optionally) temperature from MPAS netCDF along a transect.
+@author: trevorhillebrand
+"""
+
+import numpy as np
+import csv
+from netCDF4 import Dataset
+from optparse import OptionParser
```
matthewhoffman, line 12: Given that optparse is deprecated, we should probably use argparse for new scripts.

landice/output_processing_li/plot_transect.py
```diff
@@ -0,0 +1,214 @@
...
+# -*- coding: utf-8 -*-
+"""
+Plot bed and surface topography, surface speed, and
+(optionally) temperature from MPAS netCDF along a transect.
+@author: trevorhillebrand
+"""
+
+import numpy as np
+import csv
+from netCDF4 import Dataset
```
matthewhoffman, line 11: Ideally we would use xarray to allow support for times from different files in a multi-file dataset.  But I'm happy to save that work for the first person who needs the functionality rather than making this review process take a bunch of time to make a perfect script.

landice/output_processing_li/plot_transect.py
```diff
@@ -0,0 +1,214 @@
...
+li_mask_ValueDynamicIce = 2
+cellMask = dataset.variables['cellMask'][:]
+cellMask_dynamicIce = (cellMask & li_mask_ValueDynamicIce) // li_mask_ValueDynamicIce
+# only take thickness of dynamic ice
+thk = dataset.variables["thickness"][:] * cellMask_dynamicIce
+# Include speed on non-dynamic ice to avoid interpolation artifacts.
+if "surfaceSpeed" in dataset.variables.keys():
+    speed = dataset.variables["surfaceSpeed"][:] * 3600. * 24. * 365.
+else:
+    speed = np.sqrt(dataset.variables["uReconstructX"][:,:,0]**2. +
```
matthewhoffman, line 89: Do you want to check if uReconstructX/Y are available?  It seems the script won't work if there is no velocity data in the output file.  I'd be fine committing it with the requirement that some form of velocity data are present (which is usually going to be the case), and saving changing that for if such a use case comes up.

landice/output_processing_li/plot_transect.py
```diff
@@ -0,0 +1,214 @@
...
+# Include speed on non-dynamic ice to avoid interpolation artifacts.
+if "surfaceSpeed" in dataset.variables.keys():
+    speed = dataset.variables["surfaceSpeed"][:] * 3600. * 24. * 365.
+else:
+    speed = np.sqrt(dataset.variables["uReconstructX"][:,:,0]**2. + 
+                    dataset.variables["uReconstructY"][:,:,0]**2.)
+    speed *= 3600. * 24. * 365.  # convert from m/s to m/yr
+
+if options.interp_temp:
+    temperature = dataset.variables['temperature'][:]
```
matthewhoffman, line 94: Are you sure you need to read in the entire temperature array?  (Same goes for speed and thickness arrays above.)  For small files this is fine, but it might get inefficient for large files.  From what I'm seeing so far below, I think you can get away with dropping the `[:]` here and only accessing the data for the requested time levels when needed below.

landice/output_processing_li/plot_transect.py
```diff
@@ -0,0 +1,214 @@
...
+            x.append(float(row[0]))
+            y.append(float(row[1]))
+    if [options.x_coords, options.y_coords] is not [None, None]:
+        print('-c and -x/-y options were both provided. Reading from ',
+              f'{options.coords_file} and ignoring -x and -y settings.')
+else:
+    x = [float(i) for i in options.x_coords.split(',')]
+    y = [float(i) for i in options.y_coords.split(',')]
+
+# increase sampling by a factor of 100
```
matthewhoffman, line 116: Do you want to make this a function of dcEdge instead of always using 100?  (Another thing that could be saved for future improvement)

landice/output_processing_li/plot_transect.py
```diff
@@ -0,0 +1,214 @@
...
+    speed = dataset.variables["surfaceSpeed"][:] * 3600. * 24. * 365.
+else:
+    speed = np.sqrt(dataset.variables["uReconstructX"][:,:,0]**2. + 
+                    dataset.variables["uReconstructY"][:,:,0]**2.)
+    speed *= 3600. * 24. * 365.  # convert from m/s to m/yr
+
+if options.interp_temp:
+    temperature = dataset.variables['temperature'][:]
+    
+bedTopo = dataset.variables["bedTopography"][0,:]
```
matthewhoffman, line 96: Add a note it's using the initial bed only.  We may be having changing bed topography routinely soon, and when that's so, this script will be useful for looking at that.  For now, simply printing a warning about this would be fine with me.

matthewhoffman (changes requested):
@trhille , this looks like a great tool to have.  I would have suggested seeing what's being  done for transects in MPAS-Analysis to avoid reinventing the wheel, but there is enough ice-sheet-specific stuff in here (and you already have this done) that I'm not going to suggest that now.  I made a number of comments and questions, but I don't want to make this a really onerous review process.

trhille:
Thanks for the review @matthewhoffman! I'll get on these changes.

>  I would have suggested seeing what's being done for transects in MPAS-Analysis to avoid reinventing the wheel.

I already had most of this lying around in a branch from a few years ago, so really the new thing was the temperature plotting.

landice/output_processing_li/plot_transect.py
trhille: `argparse` does not seem to be able to handle negative numbers in the `-x` and `-y` arguments. This seems to be a known bug that was possibly fixed a while ago, but `argparse.__version__` returns `1.1` for both `dev_compass_1.2.0-alpha.6` and `e3sm_unified_1.9.1_login`, while the latest version of argparse is 3.12.

landice/output_processing_li/plot_transect.py
trhille: Resolved in 099bf72e

landice/output_processing_li/plot_transect.py
trhille: Added in commit 48afa27.

landice/output_processing_li/plot_transect.py
trhille: Addressed in commit 1d0e288.

landice/output_processing_li/plot_transect.py
trhille: Yeah, let's leave that for future development because I don't have time to make that change now.

trhille:
@matthewhoffman, I've addressed all your comments in one form or another. The change to `argparse` rather than `optparse` is clunky enough that I'd rather leave it using `optparse`, even though it's deprecated.

landice/output_processing_li/plot_transect.py
matthewhoffman: Sounds good

landice/output_processing_li/plot_transect.py
matthewhoffman: I agree that is a big enough annoyance to stick with optparse until we're forced to update.

landice/output_processing_li/plot_transect.py
```diff
@@ -0,0 +1,220 @@
...
+            y.append(float(row[1]))
+    if [options.x_coords, options.y_coords] != [None, None]:
+        print('-c and -x/-y options were both provided. Reading from ',
+              f'{options.coords_file} and ignoring -x and -y settings.')
+else:
+    x = [float(i) for i in options.x_coords.split(',')]
+    y = [float(i) for i in options.y_coords.split(',')]
+
+# increase sampling to match highest mesh resolution
+total_distance = np.sqrt( (x[0] - x[1])**2. + (y[0] - y[1])**2.)
```
matthewhoffman, line 120: Why not use the `distance` variable you calculate on line 132?  In addition to the fact that you are already calculating it, it would handle curved transects better.  And isn't this approach just considering the distance between the first two points, but it's possible to have many points defining a transect?

trhille: Fixed in e303679.

PR #543, opened 2023-12-13
Add check that basalHeatFlux is non-negative

trhille:
Add check that `basalHeatFlux` is non-negative. If any negative values are found, throw an error and write out a message to check the source data set.


matthewhoffman (approved):
Thanks, @trhille !

PR #544, opened 2023-12-17
Fix bugs in ESMF interpolation method

matthewhoffman:
This merge fixes a number of issues with the 'esmf' method for interpolation in the MALI interpolation script:
* indexing was off by one, which made interpolation with unstructured source meshes garbage.  For structured source meshes, it made interpolation shifted by one grid cell.
* support MPAS source fields with a vertical dimension when using the 'esmf' method
* refactor to use sparse matrix multiply, which speeds up interpolation a few hundred times
* add destination mesh area normalization support.  This is necessary when using the ESMF 'conserve' method for destination cells that are only partly overlapped by the source cells.

Two issues remain for interpolating between two MPAS meshes with the 'esmf' method:
1. If the destination mesh is larger than the source mesh, those locations are filled with zeros.  The ESMF 'conserve' method does not support extrapolation, and there is no obvious solution to this issue and it would need to be handled manually on a case by case basis.
2. Some fields are only defined on subdomains (e.g. temperature is only defined where ice thickness is nonzero) and the script currently has no mechanism for masking them.  This results in garbage values getting interpolated in, e.g. temperature values near the margin will have values around 100 K because of interpolating realistic values around 250K with garbage values of 0K.  This issue applies to the barycentric method as well.  However, this situation can be worked around by performing extrapolation of the temperature field on the source mesh before doing interpolation between meshes.

This PR also includes a refactoring of create_SCRIP_file_from_planar_rectangular_grid.py that speeds it up by orders of magnitude for large meshes, as well as a minor update to define_cullMask.py

### DIFF
--- a/landice/mesh_tools_li/interpolate_to_mpasli_grid.py
+++ b/landice/mesh_tools_li/interpolate_to_mpasli_grid.py
@@ -32,6 +32,7 @@ import argparse
 import math
 from collections import OrderedDict
 import scipy.spatial
+import scipy.sparse
 import time
 from datetime import datetime
 
@@ -64,8 +65,13 @@ if args.weightFile and args.interpType == 'e':
     S = wfile.variables['S'][:]
     col = wfile.variables['col'][:]
     row = wfile.variables['row'][:]
+    n_a = len(wfile.dimensions['n_a'])
+    n_b = len(wfile.dimensions['n_b'])
     wfile.close()
     #----------------------------
+
+    # convert to sparse matrix format
+    wt_csr = scipy.sparse.coo_array((S, (row - 1, col - 1)), shape=(n_b, n_a)).tocsr()
 
 print('') # make a space in stdout before further output
 
@@ -78,15 +84,13 @@ print('') # make a space in stdout before further output
 #----------------------------
 
 def ESMF_interp(sourceField):
-    # Interpolates from the sourceField to the destinationField using ESMF weights
+  # Interpolates from the sourceField to the destinationField using ESMF weights
+  destinationField = np.zeros(xCell.shape)  # fields on cells only
   try:
-    # Initialize new field to 0 - required
-    destinationField = np.zeros(xCell.shape)  # fields on cells only
-    sourceFieldFlat = sourceField.flatten()  # Flatten source field
-    for i in range(len(row)):
-      destinationField[row[i]-1] = destinationField[row[i]-1] + S[i] * sourceFieldFlat[col[i]-1]
+    source_csr = scipy.sparse.csr_matrix(sourceField.flatten()[:, np.newaxis])
+    destinationField = wt_csr.dot(source_csr).toarray()
   except:
-     'error in ESMF_interp'
+    print('error in ESMF_interp')
   return destinationField
 
 #----------------------------
--- a/landice/mesh_tools_li/interpolate_to_mpasli_grid.py
+++ b/landice/mesh_tools_li/interpolate_to_mpasli_grid.py
@@ -67,6 +67,7 @@ if args.weightFile and args.interpType == 'e':
     row = wfile.variables['row'][:]
     n_a = len(wfile.dimensions['n_a'])
     n_b = len(wfile.dimensions['n_b'])
+    dst_frac = wfile.variables['frac_b'][:]
     wfile.close()
     #----------------------------
 
@@ -89,6 +90,10 @@ def ESMF_interp(sourceField):
   try:
     source_csr = scipy.sparse.csr_matrix(sourceField.flatten()[:, np.newaxis])
     destinationField = wt_csr.dot(source_csr).toarray().squeeze()
+    # For conserve remapping, need to normalize by destination area fraction
+    # It should be safe to do this for other methods
+    ind = np.where(dst_frac > 0.0)[0]
+    destinationField[ind] /= dst_frac[ind]
   except:
     print('error in ESMF_interp')
   return destinationField
--- a/landice/mesh_tools_li/interpolate_to_mpasli_grid.py
+++ b/landice/mesh_tools_li/interpolate_to_mpasli_grid.py
@@ -88,7 +88,7 @@ def ESMF_interp(sourceField):
   destinationField = np.zeros(xCell.shape)  # fields on cells only
   try:
     source_csr = scipy.sparse.csr_matrix(sourceField.flatten()[:, np.newaxis])
-    destinationField = wt_csr.dot(source_csr).toarray()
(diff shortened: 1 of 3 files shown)


matthewhoffman:
@trhille , I decided to leave this branch as is and ask for review now (no rush).  The issue about the conserve method filling 0s in regions of extrapolation of the source mesh is a larger problem that it is not obvious how to resolve, so I left it as a known issue for now.  But if you have ideas about appropriate functionality for that situation, I'm happy to do further work in this PR.

matthewhoffman:
@xylar , there is a commit in this PR that makes an update to `mesh_tools/create_SCRIP_files/create_SCRIP_file_from_planar_rectangular_grid.py`.  However, it is unclear to me if I should also be updating the copy of this file at `conda_package/mesh_tools/create_SCRIP_files/create_SCRIP_file_from_planar_rectangular_grid.py`.  It seems that I need to do that for compass to be able to see my changes, but when I try to commit that change, I get this error:
```
The following paths are ignored by one of your .gitignore files:
conda_package/mesh_tools
```
Can you explain to me how this is supposed to be handled?  I've never fully understood how the conda package interacts (or doesn't) with the legacy scripts.

xylar:
@matthewhoffman, when you do a local install of the `conda_package` into a conda environment, e.g. with:
```
python -m pip install conda_package
```
it makes a copy of the ocean, land_ice and other directories that contain scripts because it can't include anything in the python package that lives outside of `conda_package`.  So if you're modifying scripts outside of `conda_package`, you should run the command above again to first copy the script(s) over again and then install them into the conda environment.

Ideally, if we want to have `mpas_tools` be a "proper" python package, we would stop making scripts and switch to entry points that live within `conda_package/mpas_tools`.  But given that that's not the workflow that you all are used to, this copying over is necessary.

matthewhoffman:
@xylar , thanks for clarifying that.  I had run `python -m pip install conda_package`  but I hadn't appreciated that because of the legacy scripts, that that command would need to be rerun every time one of those scripts was modified.  Now I understand the workflow and have more incentive to move things into the conda package. :)

matthewhoffman:
# Testing

These updated scripts were used for the 2km AIS mesh generation in https://github.com/MPAS-Dev/compass/pull/750 and it worked as expected end to end.  The refactoring sped up those individual operations dramatically.

landice/mesh_tools_li/interpolate_to_mpasli_grid.py
```diff
@@ -64,9 +65,15 @@
...
     col = wfile.variables['col'][:]
     row = wfile.variables['row'][:]
+    n_a = len(wfile.dimensions['n_a'])
+    n_b = len(wfile.dimensions['n_b'])
+    dst_frac = wfile.variables['frac_b'][:]
     wfile.close()
     #----------------------------
 
+    # convert to sparse matrix format
+    wt_csr = scipy.sparse.coo_array((S, (row - 1, col - 1)), shape=(n_b, n_a)).tocsr()
```
trhille, line 75: I dislike the variable name `wt_csr`, because it requires googling to figure out what it means. Can we use a longer, more descriptive name? Also, is `wt` short for `weight`?

landice/mesh_tools_li/interpolate_to_mpasli_grid.py
```diff
@@ -78,15 +85,17 @@
...
-    # Interpolates from the sourceField to the destinationField using ESMF weights
+  # Interpolates from the sourceField to the destinationField using ESMF weights
+  destinationField = np.zeros(xCell.shape)  # fields on cells only
   try:
-    # Initialize new field to 0 - required
-    destinationField = np.zeros(xCell.shape)  # fields on cells only
-    sourceFieldFlat = sourceField.flatten()  # Flatten source field
-    for i in range(len(row)):
-      destinationField[row[i]-1] = destinationField[row[i]-1] + S[i] * sourceFieldFlat[col[i]]
+    source_csr = scipy.sparse.csr_matrix(sourceField.flatten()[:, np.newaxis])
```
trhille, line 91: Same complaint as above about the `csr` in the variable name, although here it's adjacent to the scipy method that helps explain it some.

landice/mesh_tools_li/interpolate_to_mpasli_grid.py
```diff
@@ -328,7 +337,7 @@ def interpolate_field_with_layers(MPASfieldName):
         if filetype=='cism':
            print('  Input layer {}, layer {} min/max: {} {}'.format(z, InputFieldName, InputField[z,:,:].min(), InputField[z,:,:].max()))
         elif filetype=='mpas':
-           print('  Input layer {}, layer {} min/max: {} {}'.format(z, InputFieldName, InputField[:,z].min(), InputField[z,:].max()))
```
trhille, line 331: Was this just a typo?

trhille (approved):
@matthewhoffman, this is fantastic! Thanks for making these changes, which collectively reduce the pain of mesh generation by a huge margin. I have one measly complaint about an inscrutable variable name, but otherwise this is ready to go.

xylar:
I'm about to make a new release to fix #547.  If this can be merged today, I'll wait for it.  Otherwise, it will be in the next release.

landice/mesh_tools_li/interpolate_to_mpasli_grid.py
matthewhoffman: Yeah, this was a typo that has been present from whenever this print statement was introduced.  It wasn't obviously incorrect from the output because the typo version still yielded reasonable values.

landice/mesh_tools_li/interpolate_to_mpasli_grid.py
matthewhoffman: Updated in b0e614726526305d44e62fa83111ee6e750c322a.

landice/mesh_tools_li/interpolate_to_mpasli_grid.py
matthewhoffman: Updated in b0e614726526305d44e62fa83111ee6e750c322a.

trhille:
@matthewhoffman, thanks! That looks great. This is ready to merge, as far as I'm concerned, although I see that two of the CI checks were not successful.

matthewhoffman:
@trhille , thanks for catching that.  Not sure why they didn't run, but I manually retriggered them and they all pass now.  @xylar , did you have any other feedback on this given it touches one non-MALI specific tool?

xylar (approved):
The SCRIP changes look good to me!

xylar:
I need to make an MPAS-Tools release ASAP so let me know if this can get merged today.

PR #545, opened 2024-01-03
Update AIS mesh modification scripts

matthewhoffman:
This PR includes a number of small updates to the scripts used to modify AIS initial conditions:
* fix a bug in the bulldoze script
* reduce stdout printing in bulldoze script 
* change defaults in two adjustment scripts to match what was used for the production 4km AIS mesh
* add iceMask to the interp script for mpas-to-mpas interpolation (not directly related to the other changes, but a small quality of life addition)
* add missing comma in tune_ismip6_melt_deltat.py

### DIFF
--- a/landice/mesh_tools_li/adjust_iceshelf_geometry.py
+++ b/landice/mesh_tools_li/adjust_iceshelf_geometry.py
@@ -31,7 +31,7 @@ import shutil
 
 nSmoothing = 3 # number of rounds of smoothing to apply.  3 was the best match at 8km to 30 years of thickness evolution on Ross and FRIS.  Set to 0 to disable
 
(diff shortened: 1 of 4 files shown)


matthewhoffman:
# Testing
I confirmed that the bulldozing script operated as expected on the 2km mesh after the adjustments by zooming in to the troughs that were supposed to be deeper and doing a visual confirmation the new versions looked like I remembered from when the script was implemented.

I confirmed that bathymetry & ice-shelf adjustments looked as expected, and that that interp script works with the iceMask field.

trhille (approved):
@matthewhoffman, thanks for these changes. I'm approving based on inspection and your testing.

PR #546, opened 2024-01-04
Fix docs for coastline alteration

xylar:
They were using the old (C++ based) mask creator but need to be updated to use functions in `mpas_tools.mesh.mask` now that the `mpas_tools.mesh.conversion.mask()` wrapper function uses the newer python-based mask creation and only creates region masks, not transect masks.


xylar:
@vanroekel, please confirm that the updated documentation now works for you, following up on our conversation on Slack.

vanroekel (approved):
This is consistent with the workflow I used to alter coastlines.  Thanks for the quick update @xylar!

xylar:
Thanks @vanroekel!

PR #547, opened 2024-01-08
Limit lon to -180 to 180 in lon-lat region masks

xylar:
This is needed for grids where longitude is from 0 to 360 and it is inconvenient to modify the grid before passing it to the command-line tool.

### DIFF
--- a/conda_package/mpas_tools/mesh/mask.py
+++ b/conda_package/mpas_tools/mesh/mask.py
@@ -522,6 +522,9 @@ def compute_lon_lat_region_masks(lon, lat, fcMask, logger=None, pool=None,
 
     dsMasks = xr.Dataset()
 
+    # make sure lon is between -180 and 180
+    lon = numpy.mod(lon + 180., 360.) - 180.
+
     Lon, Lat = numpy.meshgrid(lon, lat)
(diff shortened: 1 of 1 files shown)


xylar:
We have been seeing missing SOSE data in T-S diagrams and I think this is the cause.  See this example:
https://portal.nersc.gov/cfs/e3sm/xylar/analysis/E3SM-Arcticv2.1_historical0201_TS/Years1971-1990/ocean/index.html#tsDiagAntarcticRegions
Any region that only has data at negative longitudes is missing.  (Regions that cross the antimeridian are probably showing only the data from positive longitudes.)

xylar:
## Testing

I tested this as part of E3SM-Unified 1.9.2rc3.  The results are as expected (that SOSE data is available in T-S diagrams at all longitudes):
https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xasay-davis/analysis_testing/chrysalis/test_e3sm_unified_1.9.2rc3_chrysalis/main_py3.10/ocean/index.html#tsDiagAntarcticRegions
Ignore the strange color scale -- this simulation is on a very coarse mesh.

PR #549, opened 2024-01-25
jigsaw_to_netcdf: bug fix for POINT coordinates

cwsmith:
This PR fixes a bug where the third column of the `POINTS` data from a Jigsaw file is read as a spatial coordinate for 2D when it should be read as an ID, or at least ignored.  The change in this PR simply ignores it for 2D meshes.  I may submit a follow up PR to extract the ID for ongoing landice work.  

The following Jigsaw wiki describes the mesh file format: https://github.com/dengwirda/jigsaw/wiki/MSH-File-Format#point-segment

Running `jigsaw_to_netcdf` and `MpasMeshConverter.x`, with Jigsaw 1.0.0, on a 2d `EUCLIDEAN-MESH` mesh ([gis_2d_pointIds.zip](https://github.com/MPAS-Dev/MPAS-Tools/files/14058459/gis_2d_pointIds.zip)) will demonstrate the bug with `master`.  Should I add this mesh to a test data set?

All the compass `mesh_gen` tests are passing with this change.  The testing setup and test output is below.  As a sanity check I've also included the output of `pip show mpas_tools` that indicates that my local mpas_tools is used within the testing environment created with `compass setup`.

### compass test setup
```
(cmd)(mpasMain) cwsmith@checkers: /space/cwsmith/compassLandice/main (main)$ compass setup -w /space/cwsmith/compassLandice/mesh_gen_compass_main -f checkers.cfg -n 0 43 44 92 93 99
Setting up test cases:
  landice/antarctica/mesh_gen
  landice/greenland/mesh_gen
  landice/humboldt/mesh_gen
Downloading humboldt_1km_2020_04_20.epsg3413.icesheetonly.nc (42.5MiB)
  to /space/cwsmith/compassLandice/main/mpas-albany-landice
100% |#########################################################################################################################################################################################################################################################| Time:  0:00:02
  humboldt_1km_2020_04_20.epsg3413.icesheetonly.nc done.
  landice/kangerlussuaq/mesh_gen
Downloading greenland_8km_2020_04_20.epsg3413.nc (4.3MiB)
  to /space/cwsmith/compassLandice/main/mpas-albany-landice
100% |#########################################################################################################################################################################################################################################################| Time:  0:00:00
  greenland_8km_2020_04_20.epsg3413.nc done.
  landice/koge_bugt_s/mesh_gen
  landice/thwaites/mesh_gen
Downloading antarctica_1km_2020_10_20_ASE.nc (559.3MiB)
[...]

PR #550, opened 2024-01-29
setup: always copy subdirs ocean, landice,...

cwsmith:
If changes are made to the scripts in these directories running `python -m pip install -e conda_package` will not update them.  This change removes the directory existence check so that they are always copied.

Thanks to @xylar for the pointer on the fix for this.

closes #549

### DIFF
--- a/conda_package/setup.py
+++ b/conda_package/setup.py
@@ -34,8 +34,10 @@ version = re.search(r'{}\s*=\s*[(]([^)]*)[)]'.format('__version_info__'),
 os.chdir(here)
 
 for path in ['ocean', 'landice', 'visualization', 'mesh_tools']:
-    if not os.path.exists(path):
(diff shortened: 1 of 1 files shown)


xylar (approved):
@cwsmith, this looks great

I tested it out with:
```
conda create -y -n test mpas_tools
conda activate test
python -m pip install -e conda_package/
```
Then, I modified `landice/mesh_tools_li/tune_ismip6_melt_deltat.py`, ran:
```
python -m pip install -e conda_package/
```
again and verified that `conda_package/landice/mesh_tools_li/tune_ismip6_melt_deltat.py` contained the changes.

xylar:
@cwsmith, this will appear in the next release but likely won't make it into Compass immediately because I think we pin a specific version.  Hopefully, that won't affect you because presumably your development branches will always start from `master`.

PR #551, opened 2024-01-31
Ignore NaNs when checking for negative basalHeatFlux

trhille:
Ignore NaNs when checking for negative basalHeatFlux by using np.nanmin() instead of just min(). This will not throw an assertion error when NaNs are present in the interpolated basalHeatFlux field. NaNs in datasets are a separate issue from sign conventions, so will be dealt with separately. This merge fixes a bug introduced in #543.


matthewhoffman (approved):
Tested this in https://github.com/MPAS-Dev/compass/pull/767 and the AIS mesh gen tests passed.

matthewhoffman (approved):

PR #552, opened 2024-02-14
Make some variables optional in MpasCellCuller.x

xylar:
We want to be able to run the cell culler even on meshes (e.g. in MPAS-Seaice restart files) without `meshDensity`, `weightsOnEdge`, and `angleEdge` as variables.

I have made some updates to the configurations used to make local package builds and run CI so they match the package versions that conda-forge is using.  I include this here because I used it in testing these changes.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -38,7 +38,7 @@ jobs:
   - bash: |
       eval "$(conda shell.bash hook)"
       conda activate build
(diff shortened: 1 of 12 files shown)


xylar:
## Testing

I was able to successfully run the `simple_seaice_partitions` tool pointing to an MPAS-Seaice restart file as the mesh file, whereas I got errors related to the 3 missing variables without these fixes.

PR #559, opened 2024-03-22
Add south polar stereographic projection of a sphere

hollyhan:
Previously, the south polar projection onto the planar MALI mesh was done based on an ellipsoidal Earth. However, using the MALI mesh using this projection in coupled MALI and Sea Level Model simulations resulted in inconsistency in post-simulation ice mass calculation over Antarctica because the SLM grid assumes spherical global grid in its calculation. To address this issue, while not a fundamental fix in the meshing process, an interim solution was made to project the southern polar region onto the MALI planar mesh assuming a sphere so that it is consistent with the SLM; this PR adds the new projection.


hollyhan:
Testing: 

Plot showing grounded idealized, grounded ice mass over Antarctica on the MALI domain and the SLM domain. Black dashed line shows ice mass integrated over the planar MALI domain where the projection assumes a spherical earth, and black solid line shows the former scaled back onto sphere by applying the scaling ('k-') factor of a sphere, and pink dashed-dotted line represents the grounded ice mass in its native spherical grid. 
image

xylar (approved):
Looks right to me.

matthewhoffman (approved):
Looks good, thanks, Holly

PR #560, opened 2024-03-26
Add a grow mask to constrain flood fill mask

xylar:
This merge adds a `daGrow` argument to `compute_mpas_flood_fill_mask()` that can be used to constrain where we are allowed to flood.  Previously, this function was only applied on culled meshes and was used to make sure all cells in the culled mesh were connected to each other.  With this change, the function can be used for more general flood fills.


xylar:
## Testing

The `daGrow` argument has been used in Compass to make sure the land-ice mask does not have holes (by flood filling an "ocean" mask that is the inverse of the land-ice mask) and then to ensure that all land-ice cells are connected to each other (by flood filling the land-ice mask from the south pole).  Tests were performed on IcoswISC meshes at both 240 km and 30 km resolution.

PR #563, opened 2024-04-12
ISMIP6 2500 Projections Forcing

andrewdnolan:
A standalone script to generate ISMIP6 2500 projects forcing files, from existing 2300 forcing files.


matthewhoffman (approved):
Looks great based on inspection, and I ran it for the 2km mesh and got the desired results.

PR #564, opened 2024-04-17
add adjust_thickness_at_false_ocean_cells.py

alexolinhager:
Adds a script that adjusts the ice thickness at cells inland from the grounding line that are falsely identified as part of the ocean (based on ocean density floating criteria), in order to re-designate these cells as grounded.


alexolinhager:
Before and after screenshots of false ocean cells near Ross Ice Shelf. Black is floating, red is grounded
Screenshot 2024-04-17 at 3 55 31 PM

Zoomed in:
Screenshot 2024-04-17 at 3 55 53 PM

matthewhoffman (approved):
@alexolinhager , looks great, thanks for adding this script!

PR #566, opened 2024-04-23
Add plotting script for thermal forcing and surface mass balance

trhille:
This merge adds a script that plots thermal forcing and surface mass balance time series. Currently the thermal forcing-related fields are hard-coded to ISMIP6 variable names, but that could be changed in the future.  The user must specify files from which to plot thermal forcing and SMB; a mesh file containing coordinates, ice geometry, etc; and optionally a regions file that contains the `regionCellMasks` field.
There are many (maybe too many) user-defined options for plotting:
1. The option to plot at (i) a specified depth (or average over a depth range) using the `-d` argument, (ii) to interpolate the thermal forcing field to the seafloor (`--seafloor`), ice-shelf base (`--shelf_base`), or (iii) average over the full ice-shelf depth range (`--shelf_range`).
2. The option to specify a csv file containing x and y coordinates for thermal forcing using the `-c` flag, or to specify coordinates in the command using `-x` and `-y` as comma-separated lists.
3. The option to set start and end time indices using `--start_time` and `--end_time`.
4. The option to take a random sample of point from the thermal forcing coordinates using `--n_samples`. This is just for prototyping, and the random samples are not currently reproducible.
5. The option to average thermal forcing over all cells using the `--average` flag. If this flag is not set, a separate time series will be plotted for every cell, which will get messy when plotting more than a few cells. Note that averaging is not currently weighted by cell area.
6. The option to save .png and .pdf files using the `--save` flag.
7. The option to plot SMB for a specified region using the `-n` flag, which corresponds to the region number in `regionCellMasks`.

### DIFF
--- a/landice/output_processing_li/plot_thermal_forcing_time_series.py
+++ b/landice/output_processing_li/plot_thermal_forcing_time_series.py
@@ -53,6 +53,9 @@ parser.add_option("--seafloor", dest="plot_seafloor_thermal_forcing",
 parser.add_option("--shelf_base", dest="plot_shelf_base_thermal_forcing",
                   action="store_true",
                   help="plot thermal forcing at the base of floating ice, instead of at specific depth")
+parser.add_option("--shelf_range", dest="plot_shelf_depth_range_thermal_forcing",
+                  action="store_true",
+                  help="plot average thermal forcing over the whole depth range spanned by the ice shelf base")
 parser.add_option("--average", dest="plot_average_thermal_forcing",
                   action="store_true", help='Whether to plot average' \
                   ' thermal forcing across all coordinates provided.')
@@ -148,6 +151,13 @@ def interp_and_plot_tf(tf_file, plot_ax):
     elif options.plot_shelf_base_thermal_forcing:
         # Assume this is floating ice
         depth = -1. * rhoi / rhosw * thk[0, nearest_cell]
+    elif options.plot_shelf_depth_range_thermal_forcing:
+        depth = np.zeros(2)
+        depth[0] = np.max( (-1. * rhoi / rhosw * thk[0, nearest_cell])
+                           [thk[0, nearest_cell] > 10.] )  # ignore very thin ice
+        depth[1] = np.min( (-1. * rhoi / rhosw * thk[0, nearest_cell])
+                           [thk[0, nearest_cell] > 10.] )
+        plot_depth_range = True
 
     # Clip depth to within the range of the TF data
     depth[depth > np.max(z)] = np.max(z)
--- a/landice/output_processing_li/plot_thermal_forcing_time_series.py
+++ b/landice/output_processing_li/plot_thermal_forcing_time_series.py
@@ -17,11 +17,16 @@ from matplotlib.pyplot import cm
 
 
 parser = OptionParser(description='Plot transect from MPAS netCDF')
-parser.add_option("-f", "--file", dest="thermal_forcing_file",
+parser.add_option("-t", "--tf", dest="thermal_forcing_file",
                   help="List of MPAS netCDF files that contains the ismip6shelfMelt_3dThermalForcing" \
                        " field and zOcean variable. Comma-separated, no spaces.")
+parser.add_option("-s", "--smb", dest="smb_file",
+                  help="List of MPAS netCDF files that contains the sfcMassBal" \
+                       " field. Comma-separated, no spaces.")
 parser.add_option("-m", "--mesh", dest="mesh_file",
                   help="the MPAS netCDF file that contains the mesh variable, as well as thickness and bedTopography")
+parser.add_option("-r", "--regions", dest="regions_file", default=None,
+                  help="the MPAS netCDF file that contains the region masks")
 parser.add_option("--start_time", dest="start_time", default="0",
                   help="beginning of time range to plot")
 parser.add_option("--end_time", dest="end_time", default="-1",
@@ -40,6 +45,8 @@ parser.add_option('-d', dest='depth', default=None,
                         ' interpolation to determine the thermal forcing' \
                         ' at that depth. If two values are given, the script' \
                         ' will provide the average over that depth range.')
+parser.add_option('-n', dest='region_number', default=None,
+                  help='Region number to plot. If None, use entire domain.')
 parser.add_option("--seafloor", dest="plot_seafloor_thermal_forcing",
                   action="store_true",
                   help="plot thermal forcing at the seafloor, instead of at specific depth")
@@ -51,7 +58,7 @@ parser.add_option("--average", dest="plot_average_thermal_forcing",
                   ' thermal forcing across all coordinates provided.')
 parser.add_option("--n_samples", dest="n_samples", default=None,
                   help="Number of random samples to take from provided coordinates.")
-parser.add_option("-s", dest="save_filename", default=None,
+parser.add_option("--save", dest="save_filename", default=None,
                    help="File to save figure to.")
 
 options, args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


trhille:
Example plot, created using the following command:
```
python plot_forcing_time_series.py -t /global/cfs/cdirs/fanssie/MALI_projects/ISMIP6-2300/forcing/ais_mesh_4to20km_res/tier1_experiments/expAE02/AIS_4to20km_r01_20220907_TF_CCSM4_RCP85_2300.nc,/global/cfs/cdirs/fanssie/MALI_projects/ISMIP6-2300/forcing/ais_mesh_4to20km_res/tier1_experiments/expAE03/AIS_4to20km_r01_20220907_TF_HadGEM2-ES_RCP85_2300.nc,/global/cfs/cdirs/fanssie/MALI_projects/ISMIP6-2300/forcing/ais_mesh_4to20km_res/tier1_experiments/expAE04/AIS_4to20km_r01_20220907_TF_CESM2-WACCM_SSP585_2300.nc,/global/cfs/cdirs/fanssie/MALI_projects/ISMIP6-2300/forcing/ais_mesh_4to20km_res/tier1_experiments/expAE05/AIS_4to20km_r01_20220907_TF_UKESM1-0-LL_SSP585_2300.nc,/global/cfs/cdirs/fanssie/MALI_projects/ISMIP6-2300/forcing/ais_mesh_4to20km_res/tier2_experiments/expAE10/AIS_4to20km_TF_UKESM1-0-LL_SSP126_2300.nc -s /global/cfs/cdirs/fanssie/MALI_projects/ISMIP6-2300/forcing/ais_mesh_4to20km_res/tier1_experiments/expAE02/AIS_4to20km_r01_20220907_SMB_CCSM4_RCP85_2300.nc,/global/cfs/cdirs/fanssie/MALI_projects/ISMIP6-2300/forcing/ais_mesh_4to20km_res/tier1_experiments/expAE03/AIS_4to20km_r01_20220907_SMB_HadGEM2-ES_RCP85_2300.nc,/global/cfs/cdirs/fanssie/MALI_projects/ISMIP6-2300/forcing/ais_mesh_4to20km_res/tier1_experiments/expAE04/AIS_4to20km_r01_20220907_SMB_CESM2-WACCM_SSP585_2300.nc,/global/cfs/cdirs/fanssie/MALI_projects/ISMIP6-2300/forcing/ais_mesh_4to20km_res/tier1_experiments/expAE05/AIS_4to20km_r01_20220907_smb_UKESM1-0-LL_SSP585_2300.nc,/global/cfs/cdirs/fanssie/MALI_projects/ISMIP6-2300/forcing/ais_mesh_4to20km_res/tier2_experiments/expAE10/AIS_4to20km_SMB_UKESM1-0-LL_SSP126_2300.nc -m /global/cfs/cdirs/fanssie/MALI_projects/ISMIP6-2300/initial_conditions/AIS_4to20km_20230105/relaxation_0TGmelt_10yr/relaxed_10yrs_4km.nc -r /global/cfs/cdirs/fanssie/MALI_projects/ISMIP6-2300/initial_conditions/AIS_4to20km_20230105/AIS_4to20km_r01_20220907.regionMask_ismip6.nc -c /global/cfs/cdirs/fanssie/MALI_projects/Amery_UQ/amery.csv -n 2 --shelf_base --average --start_time 5 --end_time -1 --save /pscratch/sd/t/trhille/Amery_TF_SMB
```
Amery_TF_SMB

landice/output_processing_li/plot_forcing_time_series.py
```diff
@@ -0,0 +1,272 @@
...
+                        using a csv file. Comma-separated, no spaces.')
+parser.add_option('-y', dest='y_coords', default=None,
+                  help='List of y coordinates of transect if not \
+                        using a csv file. Comma-separated, no spaces.')
+parser.add_option('-d', dest='depth', default=None,
+                  help='Depth in meters at which to plot thermal forcing.' \
+                        ' If a single value, the script will use linear 1d' \
+                        ' interpolation to determine the thermal forcing' \
+                        ' at that depth. If two values are given, the script' \
+                        ' will provide the average over that depth range.')
```
matthewhoffman, line 47:
```suggestion
                        ' at that depth. If two comma-delimited values are given, the script' \
                        ' will provide the average over that depth range.')
```

landice/output_processing_li/plot_forcing_time_series.py
```diff
@@ -0,0 +1,272 @@
...
+    if options.depth is not None:
+        depth = np.array([float(i) for i in options.depth.split(',')])
+        if len(depth) == 2:
+           plot_depth_range = True
+
+    if options.plot_seafloor_thermal_forcing:
+        depth = bed[0, nearest_cell]
+    elif options.plot_shelf_base_thermal_forcing:
+        # Assume this is floating ice
+        depth = -1. * rhoi / rhosw * thk[0, nearest_cell]
```
matthewhoffman, line 153: Do you want to do this?  Because TF is defined everywhere, it would be easy to inadvertently include values inside grounded ice.  Maybe that's desired, though generally probably not?

landice/output_processing_li/plot_forcing_time_series.py
```diff
@@ -0,0 +1,272 @@
...
+parser.add_option("--end_time", dest="end_time", default="-1",
+                  help="end of time range to plot")
+parser.add_option('-c', dest='coords_file', default=None,
+                  help='CSV file containing x in first column, y in second. No header.')
+parser.add_option('-x', dest='x_coords', default=None,
+                  help='List of x coordinates of transect if not \
+                        using a csv file. Comma-separated, no spaces.')
+parser.add_option('-y', dest='y_coords', default=None,
+                  help='List of y coordinates of transect if not \
+                        using a csv file. Comma-separated, no spaces.')
```
matthewhoffman, line 41: From what I understand of the code below, it looks like it is necessary to specify x,y locations in one of these two ways.  Is that right?  Is there not a way to just use all data in a region, say?  I'm not asking for you to generalize that behavior if it doesn't already exist, but clarify, probably in the main doc strings, how this works.  Also that all the depth options apply over just the horizontal locations as defined by these options, i.e. plot_shelf_base_thermal_forcing is not the entire shelf base but the shelf base at the x,y locations determined by these horizontal options. (if I follow correctly)

landice/output_processing_li/plot_forcing_time_series.py
```diff
@@ -0,0 +1,272 @@
...
+parser.add_option('-d', dest='depth', default=None,
+                  help='Depth in meters at which to plot thermal forcing.' \
+                        ' If a single value, the script will use linear 1d' \
+                        ' interpolation to determine the thermal forcing' \
+                        ' at that depth. If two values are given, the script' \
+                        ' will provide the average over that depth range.')
+parser.add_option('-n', dest='region_number', default=None,
+                  help='Region number to plot. If None, use entire domain.')
```
matthewhoffman, line 49: It looks like the region only applies to the SMB plot - is that intended?

matthewhoffman (changes requested):
@trhille , thanks for making a PR to capture this script.  I have some cleanup suggestions and a couple questions about the functionality.

landice/output_processing_li/plot_forcing_time_series.py
trhille: That's correct. I do want to generalize it, but it wasn't necessary for the Lambert paper figure (although it might have made things easier if I had started with that). I can add a comment for the region is just SMB for now.

landice/output_processing_li/plot_forcing_time_series.py
trhille: There's not currently a way to use all data in a region, but that would be good to add. I'll update the doc strings.

landice/output_processing_li/plot_forcing_time_series.py
trhille: Because this only plots TF at user-specified locations and the concept of the shelf base is meaningless for grounded ice, I don't think this is a problem. But I could add an assertion that depth >= bed.

Issue #565, opened 2024-04-22
Wrong `areaTriangle` values in mesh created with `MpasMeshConverter.x` 


favba:
I'm using the `MpasMeshConverter.x` tool to create planar periodic meshes by feeding cells and vertices position, together with the connectivity, and I noticed that the area of the dual cells (triangles) computed by the tool is not correct.

I'll provide a MWE when I have the time. Meanwhile, this is the ncdump of the mesh created where I noticed this issue 
[mesh_distorted_dump.txt](https://github.com/MPAS-Dev/MPAS-Tools/files/15062879/mesh_distorted_dump.txt)
The mesh was created based on the `xCell`,`yCell`,`zCell`,`xVertex`,`yVertex`,`zVertex`, `cellsOnVertex`  and `meshDensity` info only. The rest of the data was computed by `MpasMeshConverter.x`.
I haven't checked yet if the other areas are correct.

xylar:
@favba, this is much appreciated.  It may relate to https://github.com/MPAS-Dev/MPAS-Tools/issues/289.

xylar:
Just by inspection, I don't see anything wrong with the computation of areas so this will need some deeper debugging.

favba:
An important observation: The meshes I'm generating are purposely inhomogeneous. I first generate a homogeneous mesh with `planar_hex`, then I disturb the voronoi cells generator points so I don't have a perfectly homogeneous mesh anymore, then recompute the triangles circumcenters based on the new Voronoi Cells position.
Maybe the tool is assuming a homogeneous mesh?

I also noticed that
 1) while each triangle area is incorrect, the sum of all triangle areas still gives the correct result.
 2) The Voronoi Cells areas are correct
 3) The kiteAreasOnVertex are also incorrect (but consistent with the incorrect triangle area)

Here is a small mesh input where I see the issue:

```
netcdf test_mesh_distorted_incomplete {
dimensions:
	nCells = 16 ;
	nEdges = 48 ;
	nVertices = 32 ;
	maxEdges = 6 ;
	TWO = 2 ;
	maxEdges2 = 12 ;
	vertexDegree = 3 ;
variables:
	double xCell(nCells) ;
	double yCell(nCells) ;
	double zCell(nCells) ;
	double xVertex(nVertices) ;
	double yVertex(nVertices) ;
	double zVertex(nVertices) ;
	int cellsOnVertex(nVertices, vertexDegree) ;
[...]

PR #567, opened 2024-04-29
Fix `planarIntersect` formula in MPAS mesh converter

mwarusz:
The new formula is from: https://en.wikipedia.org/wiki/Line%E2%80%93line_intersection#Given_two_points_on_each_line

### DIFF
--- a/mesh_tools/mesh_conversion_tools/pnt.h
+++ b/mesh_tools/mesh_conversion_tools/pnt.h
@@ -376,18 +376,27 @@ pnt planarIntersect(const pnt &c1, const pnt &c2, const pnt &v1, const pnt &v2){
 	 * planarIntersect is intended to compute the point of intersection
 	 *    of the lines c1-c2 and v1-v2 in a plane.
 	 */
-	double alpha_numerator, beta_numerator, denom;
-	double alpha, beta;
+	double cx_numerator, cy_numerator, denom;
 	pnt c;
 
-	alpha_numerator = (v2.x - v1.x) * (c1.y - v2.y) - (v2.y - v1.y) * (c1.x - v2.x);
-//	beta_numerator = (c1.x - v2.x) * (c2.y - c1.y) - (c1.y - v2.y) * (c2.x - c1.x);
-	denom = (c2.x - c1.x) * (v2.y - v1.y) - (c2.y - c1.y) * (v2.x - v1.x);
+	double x1 = v1.x;
+	double y1 = v1.y;
 
-	alpha = alpha_numerator / denom;
-//	beta = beta_numerator / demon;
+	double x2 = v2.x;
+	double y2 = v2.y;
 
-	c = alpha * (v2 - v1) + v1;
+	double x3 = c1.x;
+	double y3 = c1.y;
+
+	double x4 = c2.x;
+	double y4 = c2.y;
+
+	cx_numerator = (x1 * y2 - y1 * x2) * (x3 - x4) - (x1 - x2) * (x3 * y4 - y3 * x4);
+	cy_numerator = (x1 * y2 - y1 * x2) * (y3 - y4) - (y1 - y2) * (x3 * y4 - y3 * x4);
+	denom = (x1 - x2) * (y3 - y4) - (y1 - y2) * (x3 - x4);
+
+	c.x = cx_numerator / denom;
+	c.y = cy_numerator / denom;
 
 	return c;
 }/*}}}*/
--- a/mesh_tools/mesh_conversion_tools_netcdf_c/pnt.h
+++ b/mesh_tools/mesh_conversion_tools_netcdf_c/pnt.h
@@ -376,18 +376,27 @@ pnt planarIntersect(const pnt &c1, const pnt &c2, const pnt &v1, const pnt &v2){
      * planarIntersect is intended to compute the point of intersection
      *    of the lines c1-c2 and v1-v2 in a plane.
      */
-    double alpha_numerator, beta_numerator, denom;
-    double alpha, beta;
+    double cx_numerator, cy_numerator, denom;
     pnt c;
 
-    alpha_numerator = (v2.x - v1.x) * (c1.y - v2.y) - (v2.y - v1.y) * (c1.x - v2.x);
-//    beta_numerator = (c1.x - v2.x) * (c2.y - c1.y) - (c1.y - v2.y) * (c2.x - c1.x);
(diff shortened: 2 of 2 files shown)


xylar:
@mwarusz, have you checked whether a similar fix is needed for spherical meshes?

xylar:
A question to @mwarusz and reviewers: is it not the case the correct edge location halfway between vertices (see https://github.com/MPAS-Dev/MPAS-Tools/issues/565#issuecomment-2082590736) rather than at this intersection?

xylar:
Finally, currently the edge location for edges with only one neighboring cell are halfway between vertices:
https://github.com/MPAS-Dev/MPAS-Tools/blob/d29c4f218dbd9a9ec4b3bc5de8e797fb1c0e1d8a/mesh_tools/mesh_conversion_tools_netcdf_c/mpas_mesh_converter.cpp#L1278
I don't see a good alternative to this but wanted to bring it up.

favba:
> A question to @mwarusz and reviewers: is it not the case the correct edge location halfway between vertices (see [#565 (comment)](https://github.com/MPAS-Dev/MPAS-Tools/issues/565#issuecomment-2082590736)) rather than at this intersection?
> 
> Perhaps the intersection formula might be correct but even if so, is it actually the case that we want the edge to be located halfway between cell centers along a great circle rather than at the intersection?

@xylar , The mesh specification is indeed a bit confusing regarding that.
In the second paragraph of chapter 2 it says:
```
As shown in Figure 2.1, a line segment that connects two primal mesh cell centers is uniquely
associated with a line segment that connects two dual mesh cell centers. We assume that these
two line segments cross and the point of intersection is labeled as x_e .
```
Which is what me and @mwarusz have been suggesting.

But later, in the first paragraph of Chapter 3 it also says:
```
In MPAS, cells
are nominally located at the Voronoi generating points, which, for centroidal Voronoi tessellations,
are the mass centoids of the Voronoi cells with respect to a density function, and edges are nominally
located at the midpoints of edges.
```
Which suggests your point of view. But I take that that "nominally" in the text means it isn't truly there...

I'll also note that my advisor published a [paper](https://www.sciencedirect.com/science/article/pii/S0021999116000267) in 2016 where, one of the things analyzed, is precisely the current edge positioning of MPAS (at intersections) vs a modified method that uses the edge midpoint.

mwarusz:
> @mwarusz, have you checked whether a similar fix is needed for spherical meshes?

I haven't checked the formula, but it seems to be doing the right thing based on numerical testing. Note that the current behavior of the spherical version and the planar version on distorted meshes is opposite with respect to edge point placement. 

> A question to @mwarusz and reviewers: is it not the case the correct edge location halfway between vertices (see #565 (comment)) rather than at this intersection?

> Perhaps the intersection formula might be correct but even if so, is it actually the case that we want the edge to be located halfway between cell centers along a great circle rather than at the intersection?

I believe this recent paper https://www.sciencedirect.com/science/article/abs/pii/S146350032100161X
from your group discusses the difference between the two approaches and concludes that what this PR does is better. Let's wait to hear what @mark-petersen has to say.

xylar:
@matthewhoffman and @trhille, MALI is likely to be the component most affected by this change, since you use irregular, planar meshes more often than the other components do.

xylar:
I tested this on our Compass ocean `pr` test suite.  It isn't bit-for-bit with the version using the previous, incorrect intersection calculation but differences were at machine precision, which is what I would have expected.  All of our planar meshes are regular so the change in edge position would be expected to be very small.

xylar:
I will make an `mpas_tools` release candidate with this fix in it to help others test it.

xylar:
A package with this change can be installed with:
```
conda create -y -n test -c conda-forge/label/mpas_tools_dev mpas_tools=0.34.0rc1
```
or in an existing environment with:
```
conda install -y -c conda-forge/label/mpas_tools_dev mpas_tools=0.34.0rc1
```

mark-petersen (approved):
Thanks. I agree with these changes based on @xylar's explanation and results.

xylar:
## More testing

I ran @favba's mesh through the converter with this update.

Before the fix, the edges and cells around the 13th vertex looked like this:
13th_vertex

With this fix, it looks as expected:
13th_vertex_fixed

xylar (approved):
Approved based on my compass testing and plotting.

mgduda (approved):
For MPAS-Atmsophere, we use the TRiSK placement of edge locations (i.e., at the midpoint of triangle faces), consistent with the changes in this PR.

proteanplanet (approved):
This is to accept the change on the planar mesh, and to confirm that for spherical meshes, the edge point falls at the intersection of the great circles between vertices and cell centers.  I have confirmed this graphically for the MPAS-Ocean/SeaIce EC30to60E2r2 mesh, centered here for the Pacific 14.5˚N 166.4˚E on a 0.75˚ horizon:

MPAS_OceanIce_AGU_EC30to60E2r2_zoom_mesh_51

Mesh lines on this plot are great circles at 1 NM resolution.

xylar:
@proteanplanet, thank you so much, that's reassuring!

matthewhoffman:
Based on the discussion and inspection of the changes, this is convincing.  @trhille  and I will test this with a few of our mesh generation workflows before approving to ensure it doesn't generate any surprises for us.

matthewhoffman (approved):
I ran a few of our compass mesh generation test cases for nonuniform meshes generated via JIGSAW using the rc mpas-tools package that @xylar conveniently set up.  

As expected, all meshes generated without incident and look identical in Paraview, even with the vtk converter.  At first I expected to see very slightly different edge locations, but I believe the vtk polygons don't actually use xEdge, yEdge, and those are the only fields modified by this PR.  I did confirm that xEdge, yEdge differ slightly before and after this PR, and that xCell, yCell, xVertex, yVertex, dcEdge, and dvEdge all are identical before and after.  Given that we don't use xEdge,yEdge anywhere in MALI (other than the mpas_li_velocity_simple.F module which hasn't been used in years), I don't anticipate this PR to actually impact simulations in any way.

So to summarize, no impact on MALI simulation, no impact on common visualization tools, and clearly a correct change - I'm happy to approve and let this move forward.

xylar:
@matthewhoffman, that's wonderful! I expected MALI to be impacted.  If not, I think we can merge this without concerns.

@mwarusz, I really appreciate your diligent and thoughtful work on this.

@favba, I deeply appreciate you bringing this up and engaging with us long enough for us to both appreciate that it was a real, significant problem and to find a solution we're happy with.

xylar:
I will make a new MPAS-Tools release soon with this fix.  I just need to check if anything else should be included.

PR #568, opened 2024-05-06
add interpolate_mpasOcean_to_mali.py

alexolinhager:
Creates new script that interpolates MPAS-Ocean thermal forcing and sub ice shelf melt rates to a MALI mesh for one way coupling. Still a work in progress at this point


alexolinhager:
@matthewhoffman, opening this now so you can get a sense of the script so far. I added "To Do" notes where needed

matthewhoffman:
# Verification testing

## melt
As a reference for the melt field, I time averaged one year of SORRM monthly output files in one directory:
`ncra -v timeMonthly_avg_landIceFreshwaterFlux * avg_FWF_2000.nc`
And then looked at the field in Paraview with:
`timeMonthly_avg_landIceFreshwaterFlux/910*3.15e7`
image

I compare this against the output of the script applied to that same set of files.  There is a sign convention error that I will push a fix for, but otherwise results look the same (`floatingBasalMassBal/910*3.15e7`  - note there should be a negative sign there):
image

So I think this verifies both the time averaging code and the grid remapping code.  

## TF

Top layer of thermal forcing in the output: 
image

23rd layer:
image

While the spatial patterns look appropriate for the melt field, the values seem off, with most of the continent having negative values.  Also, at shallow layers, there is some strange scalloping in the open ocean regions.  I can't think of a straightforward comparison to the MPAS-Ocean data here.  There could be issues with either the TF calculations, the vertical remapping, or both.

matthewhoffman:
Replaced by #577

PR #571, opened 2024-06-28
Support numpy 2.0

xylar:
So far, the only issue I have noticed with `numpy >= 2.0` is `numpy.string_` --> `numpy.bytes_`.

PR #572, opened 2024-06-29
Remove python 3.8

xylar:
This is necessary because the `pyamg` has dropped python 3.8 support in the same version that adds numpy 2.0 support (added here in #571).

This merge also cleans up `importlib.resources` imports since no external package is required for python >=3.9.


xylar:
In this case, I think CI testing is sufficient.

PR #574, opened 2024-07-03
Fix registering colormaps

xylar:
`plt.register_cmap()` seems to have been removed in the latest `matplotlib`.


xylar:
## Testing

I was able to run:
``` python
from mpas_tools.viz.colormaps import register_sci_viz_colormaps
register_sci_viz_colormaps()
```
with matplotlib 3.9.0, whereas it failed without this fix.

xylar:
This also seems to fix the error I was seeing in Compass.

PR #577, opened 2024-07-29
Tool for interpolating MPAS-Ocean output to MALI for forcing standalone simulations

matthewhoffman:
Script to interpolate existing MPAS-Ocean output files to a MALI mesh to be used to force standalone MALI simulations.

This script does the following:
* creates a mapping file between the two meshes
* collects information about the ocean mesh and generates time-independent mask files for open ocean and full ocean including cavities
* loops through selected years to:
* optionally calculate annual mean
* calculates thermal forcing from T&S
* remaps to the MALI mesh

Future improvements that are not addressed in this PR could include:
* improve handling of Time/xtime
* support time-evolving ocean vertical coordinate and mask


matthewhoffman:
This is a replacement to #568 after branches diverged.

irenavankova (commented):
These changes make sense and the outputs (shared on slack) look physically reasonable.

xylar:
The fix looks good to me, too.

PR #578, opened 2024-08-20
Add script for adding map of sufficiently deep ocean bahtymetry to a scrip file

matthewhoffman:
This PR adds a script for adding a mask to a scrip file needed for ocn-glc fjord coupling.    This script defines a mask of ocean cells that are deeper than a threshold. This is to be used for updating a scrip file so that it only remaps from ocean cells deeper than that threshold. The resulting mapping file can be used as an OCN2GLC_TF_SMAPNAME mapping file in E3SM. Note that config_2d_thermal_forcing_depth in the MPAS-Ocean namelist needs to match the depth used here.

This PR is needed for generating a mapping file to support E3SM PR:
https://github.com/E3SM-Ocean-Discussion/E3SM/pull/94


matthewhoffman:
@xylar , this is the script needed to create a masked mapping file for https://github.com/E3SM-Ocean-Discussion/E3SM/pull/94.  I'm completely open to renaming the file and/or moving its location.

Here are the steps for using this script in a workflow to generate a mapping file:
```
# create standard MPAS-Ocean scrip file
create_SCRIP_file_from_MPAS_mesh.py -m /lcrc/group/acme/public_html/inputdata/ocn/mpas-o/IcoswISC30E3r5/mpaso.IcoswISC30E3r5.20231120.nc -s ocean.IcoswISC30E3r5.mask.scrip.20231120.deeperThan300m.nc

# add mask for desired depth using the new script
~/mpas/MPAS-Tools2/mesh_tools/create_SCRIP_files/create_scrip_mask_for_sufficiently_deep_ocean_mapping.py -s ocean.IcoswISC30E3r5.mask.scrip.20231120.deeperThan300m.nc -m /lcrc/group/acme/public_html/inputdata/ocn/mpas-o/IcoswISC30E3r5/mpaso.IcoswISC30E3r5.20231120.nc -d 300

# create MALI scrip file
create_SCRIP_file_from_MPAS_mesh.py -m /lcrc/group/e3sm/public_html/inputdata/glc/mpasli/mpas.gis1to10kmR2/gis_1to10km_r02.20240513.nc -s gis_1to10km_r02.20240513.SCRIP.nc

# Create mapping file
ESMF_RegridWeightGen -r -s ocean.IcoswISC30E3r5.mask.scrip.20231120.deeperThan300m.nc -d gis_1to10km_r02.20240513.SCRIP.nc -m neareststod -w map_IcoswISC30E3r5_to_gis_1to10km_r02_esmfneareststod.deeperThan300m.20240820.nc

# The resulting mapping file should then be specified for OCN2GLC_TF_SMAPNAME mapping for that OCN/GLC grid combination
```

I have a directory on Chrysalis where the above steps have been run:
`/lcrc/group/e3sm/ac.mhoffman/scratch/mapping_files_for_ocn2glc_TF_mapping`

Note that I created the ocean scrip file **without** the landice mask.  The new bathymetry masking script will clobber the `grid_imask` set by the landice mask in `create_SCRIP_file_from_MPAS_mesh.py`, so it really does not matter if `create_SCRIP_file_from_MPAS_mesh` is called with `--landice` or not.  And that is the behavior we want, because the ocn/glc thermal forcing coupling should still use ocean grid cells inside ice-shelf cavities.  Note that facemelting is not applied on MALI underneath an ice shelf, but if the ice shelf were to disappear during a run, you would still want the mapping to exist.  This is sort of a minor detail that should generally not be applicable to Greenland conditions, but it will come into play for the few small ice shelves that exist there.  When we eventually move to passing a profile of TF built from multiple depths and using extrapolation on the MALI mesh, this script and these caveats will no longer apply.

xylar (approved):
Sorry for the slow response.

@matthewhoffman, this looks great to me as it is!

PR #579, opened 2024-09-05
Add Wave Mesh Tools 

erinethomas:
This PR adds all the tools required for creating WAVEWATCHIII meshes


xylar:
Impressive work, @erinethomas and @sbrus89!  For now, my suggestion would be to add a few tests to:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/conda_package/recipe/meta.yaml#L98
These can be as simple as making sure the executables are there (like the first 3) or even better running them to get a help message if that's possible.

erinethomas:
good suggestion - ill try to get those done before the end of the week

xylar:
I was able to build the conda package on Chrysalis so this looks great to me from a technical perspective.  Let me know if you want me to test actually running any of these tools.

PR #580, opened 2024-09-06
Fix `angleEdge` at culled mesh boundaries

xylar:
Before this fix, `angleEdge` was simply copied from the base mesh to the culled mesh.  With this change, `angleEdge` is flipped by pi radians if either the first cell on the given edge was removed (so the second cell is now the first cell after culling) or if only the second cell existed before (and after culling it is now the first cell).  This is because `angleEdge` on a boundary edge always corresponds to a normal vector pointing from the first (and only) cell on the edge to the middle of the edge.


xylar:
Unfortunately, this didn't fix reconstruction of a constant zonal velocity in my testing so I have more work to do.

PR #582, opened 2024-09-24
Add additional `pip install` flags in docs and conda recipe

altheaden:
In this PR I added a few flags to the `python -m pip install [-e] .` command. The flags are:
* `--no-deps` indicates that pip should not install runtime dependencies from PyPi.
* `--no-build-isolation` indicates that pip should not install build time dependencies from PyPi.


xylar (approved):
This looks great!  I was able to build the conda package with these changes.

PR #586, opened 2024-11-05
Transition CI from Azure Pipelines to GitHub Actions

altheaden:
In this PR, I've switched the CI workflows from azure pipelines to GH actions. There's a build workflow for testing code, and a docs workflow for deploying the documentation via the `gh-pages` branch. This change allows the CI for MPAS-Tools to be more in line with our CI procedures in other repos (e.g., Polaris, Mache, Compass).


altheaden:
@xylar As we discussed earlier, there is an issue with building the docs during the docs workflow. If you don't get a chance to debug it by tomorrow, I'm happy to try to figure it out then as well.

.github/workflows/build_workflow.yml
```diff
@@ -0,0 +1,85 @@
...
+          conda build -m "conda_package/ci/linux_64_python${{ matrix.python-version }}.____cpython.yaml" "conda_package/recipe"
+
+      - if: ${{ steps.skip_check.outputs.should_skip != 'true' }}
+        name: Install mpas_tools
+        run: |
+          cd conda_package
+          conda create -n mpas_tools_dev --file dev-spec.txt \
+              python=${{ matrix.python-version }}
+          conda activate mpas_tools_dev
+          python -m pip install --no-deps --no-build-isolation -vv -e .
```
xylar, line 75: I'm running this workflow on Chrysalis seeing errors on this command.  I think it needs to be run from the root of the branch, and have `.` replaced by `conda_package`.

xylar: If you run in `conda_package` it doesn't fine the `pyproject.toml`.

xylar:
@altheaden, I was able to reproduce the error you posted above.  It seems to happen for an old version of MPAS-Tools, 0.0.10, so we need to skip that version, similar to what was done in MPAS-Analysis:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/docs/conf.py#L233

I'll push a fix to the branch.

xylar:
Okay, I think I have the docs config cleaned up.  We just needed to skip two versions, 0.0.1 and 0.0.10 (not sure why), for the docs to build.  I did a bunch of other clean up in that config file, too, similar to what we did on other repos.

.github/workflows/build_workflow.yml
xylar, line 75: I deleted my previous comments because what actually need to change here is that we need to install the conda package we just built:
```suggestion
          conda create -y -n mpas_tools_dev --use-local mpas_tools \
              python=${{ matrix.python-version }} \
              sphinx sphinx-multiversion mock sphinx_rtd_theme
```

.github/workflows/docs_workflow.yml
```diff
@@ -0,0 +1,91 @@
...
+          python-version: ${{ matrix.python-version }}
+
+      - if: ${{ steps.skip_check.outputs.should_skip != 'true' }}
+        name: Install mpas_tools
+        run: |
+          cd conda_package
+          git config --global url."https://github.com/".insteadOf "git@github.com:"
+          conda create -n mpas_tools_dev --file dev-spec.txt
+          conda activate mpas_tools_dev
+          python -m pip install -vv --no-deps --no-build-isolation -e .
```
xylar, line 52:
```suggestion
          git config --global url."https://github.com/".insteadOf "git@github.com:"
          conda create -n mpas_tools_dev --file conda_package/dev-spec.txt
          conda activate mpas_tools_dev
          python -m pip install -vv --no-deps --no-build-isolation -e conda_package
```

xylar:
Thanks @altheaden!

PR #589, opened 2024-11-07
Add dependabot workflow

altheaden:
This PR adds the dependabot workflow. Dependabot checks for updates to dependencies on a weekly basis, and makes a PR when an update is available.


xylar (approved):
Looks great!

PR #593, opened 2024-11-14
Add python 3.13

xylar:


xylar:
@altheaden, could you have a quick look here and make sure I didn't miss anything?

xylar:
I'm also adding python 3.13 support on conda-forge here:
https://github.com/conda-forge/mpas_tools-feedstock/pull/114

.github/workflows/build_workflow.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           channels: conda-forge
           channel-priority: strict
           auto-update-conda: true
-          python-version: ${{ matrix.python-version }}
+          python-version: 3.12
```
xylar, line 60: Apparently `conda-build` isn't available for python 3.13 yet.  That's fine, we don't need to build with the full python matrix, we just need the package that *gets built* to use the full python matrix.

.github/workflows/build_workflow.yml
```diff
@@ -25,7 +25,8 @@ jobs:
         shell: bash -l {0}
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+      fail-fast: false
```
xylar, line 29: @altheaden, I've been seeing CI get cancelled when one job in the matrix fails.  I think this (rather than `cancel_others: false` below) might be the fix.

Could you add this on other repos when you do other updates?

.github/workflows/build_workflow.yml
```diff
@@ -68,8 +69,8 @@ jobs:
       - if: ${{ steps.skip_check.outputs.should_skip != 'true' }}
         name: Install mpas_tools
         run: |
-          conda create -y -n mpas_tools_dev --use-local mpas_tools \
-              python=${{ matrix.python-version }} \
+          conda create -y -n mpas_tools_dev -c ${CONDA_PREFIX}/conda-bld/ \
+              -c conda-forge mpas_tools python=${{ matrix.python-version }} \
```
xylar, line 73: I'm not sure why python 3.13 likes this better but I'll take it...

altheaden:
Looks good to me! If you would like me to do any testing myself, let me know. Otherwise, I think it looks fine to merge.

altheaden (approved):

xylar:
Thanks, @altheaden

PR #596, opened 2025-01-08
Move `landice` conda package scripts

xylar:
This merge aims to move 8 scrips used primarily or exclusively by MALI into the `mpas_tools` package as entry points.

The motivation for this is that we are engaging in Python package modernization on all MPAS-Dev and E3SM-Project repos, and in that process we discovered that modern packaging approaches do not support standalone scripts as the old approach did.

I have moved most of the code from the 8 scripts into modules within `mpas_tools`.  They are inside of functions, as required for entry points into a Python package, in all cases.  The original scripts are preserved as stubs that call the entry points.

The entry points cannot have the same names as the original scripts: the `.py` extention is not allowed.  I have also dropped capitalization, since changes are being made in any case.

While I have been fine with standalone scripts using loose Python code styles, code in the `mpas_tools` should conform to PEP8 to the extent feasible.  Therefore, most of the changes I have made here are related to updating the 8 scripts to be close to PEP8 compliant.


xylar:
@matthewhoffman, @trhille and @hollyhan, I'd like to discuss this with you when there's time (but not tonight -- I'm busy unfortunately).

We would have to make corresponding changes in Compass before this gets released, but I don't think that will be hard.

I've tried to be careful about the formatting and linting changes I made but it's a lot of code so mistakes are hard to rule out completely.

matthewhoffman:
@xylar , sorry for letting this sit so long. We'll address this early next week when Trevor is in town.

trhille:
Testing
-------
Using https://github.com/MPAS-Dev/MPAS-Tools/pull/596 rebased onto `MPAS-Dev/master`.
- [ ] Full integration suite
- [ ] `landice/antarctica/mesh_gen`
- [ ] `landice/crane/mesh_gen`
- [ ] `landice/humboldt/mesh_gen`
- [ ] `landice/greenland/mesh_gen`
- [ ] `landice/isunnguata_sermia/mesh_gen`
- [ ] `landice/kangerlussuaq/mesh_gen`
- [ ] `landice/koge_bugt_s/mesh_gen`

landice/mesh_tools_li/define_cullMask.py
```diff
@@ -1,227 +1,11 @@
...
-  plt.draw()
-  plt.show()
-
-f.close()
-print("cullMask generation complete.")
+from mpas_tools.landice.cull import define_cull_mask
 
 
+if __name__ == '__main__':
+    define_cull_mask()
```
matthewhoffman, line 11: @xylar, it looks like you retained the ability to call these scripts as before, but you also have added analogous entry points to these scripts (which could be called from the command line without the `.py` extension and are in the environment path).  That seems redundant - did you do this for our convenience?  Would you prefer we eliminate these pass-through scripts and that we just start using the entry-point versions?  I see in https://github.com/MPAS-Dev/compass/pull/881 that the entry points get used.

conda_package/mpas_tools/landice/set_lat_lon.py
```diff
@@ -0,0 +1,113 @@
+import sys
```
matthewhoffman, line 1: This script is not landice-specific and was not housed in a landice directory before, though landice is probably the only use-case, practically speaking. 
Should we move it to `mpas_tools/mesh` instead?

landice/mesh_tools_li/define_cullMask.py
xylar: We can certainly delete these wrapper scripts if you don't need them.  They won't be available outside of this repo.

conda_package/mpas_tools/landice/set_lat_lon.py
xylar: Okay, I can't remember why I decided to put it here but I maybe it was because landice is the only current user.

xylar: I'm happy to move it to `mpas_tools/mesh` as you suggest.

xylar:
@matthewhoffman (and @trhille), I have made the suggested changes -- removing most of the stub scripts and moving `set_lat_lon` into `mesh`.  The latter won't be reflected in the release candidate you're testing in Compass, obviously, but the entry point will have the same name, it just points to a different location.

xylar:
@trhille, @matthewhoffman, just checking in on this.  Anything I can do to help?

trhille:
@xylar, so sorry for the delay on this. I'm going to work on the remaining tests today.

trhille (approved):
Approved based on testing. Most of the mesh creation testing was done prior to the force push from [f4e9cf](https://github.com/MPAS-Dev/MPAS-Tools/commit/af4e9cf8ef5d7fb571cf235d58bc7440cdc8fff3) to [701c8d9](https://github.com/MPAS-Dev/MPAS-Tools/commit/701c8d97dfa733ec7128f32aff5ddd4f48a14e2c), but I re-ran the Thwaites mesh generation to make sure that didn't break anything. For the ismip6_forcing and ismip6_run tests I used the Thwaites_2to10km mesh in the `$CFS/fanssie/MALI_projects/thwaites_convergence` directory. I confirmed that relevant fields in forcing files were identical to those created previously for the Thwaites mesh convergence runs, and hist, ctrlAE, and expAE03 experiments were all set up and run successfully (only ran for a 30-minute debug job).

This testing also applies to https://github.com/MPAS-Dev/compass/pull/881.

xylar:
Thanks so much @trhille!

@matthewhoffman and @hollyhan, let me know if you'd like to review or if you're happy to leave this to @trhille.  I'll assume we're good if I don't hear anything by the end of the week.

mesh_tools/create_SCRIP_files/create_SCRIP_file_from_planar_rectangular_grid.py
```diff
@@ -2,174 +2,8 @@
...
-    plt.plot(grid_center_lon[:], grid_center_lat[:], 'bo')
-    plt.plot(grid_corner_lon[:], grid_corner_lat[:], 'g.')
-    plt.show()
-
-
-fin.close()
-fout.close()
-print('scrip file generation complete')
+if __name__ == '__main__':
+    main()
```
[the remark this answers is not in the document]
xylar: I left it because there's another stub script next to it that looks similar:
https://github.com/MPAS-Dev/MPAS-Tools/blob/701c8d97dfa733ec7128f32aff5ddd4f48a14e2c/mesh_tools/create_SCRIP_files/create_SCRIP_file_from_MPAS_mesh.py
I thought it was more likely that it might be needed outside of the conda package by some workflow or another.  Happy to delete it outright.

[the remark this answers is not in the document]
xylar: @matthewhoffman, I'd really like to move on this PR.  Would you like me to remove the stub or leave it?  Would you like me to remove the other stub as well?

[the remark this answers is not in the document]
xylar: @matthewhoffman, I'm going to remove both stubs in the hopes of moving this along.

xylar:
Okay, thanks @matthewhoffman and @trhille!  Merging this.

PR #597, opened 2025-01-09
Move `ocean` and `visualization` conda-package scripts

xylar:
This merge aims to move 5 scrips used primarily or exclusively by MPAS-Ocean in Compass into the `mpas_tools` package as entry points.

The motivation for this is that we are engaging in Python package modernization on all MPAS-Dev and E3SM-Project repos, and in that process we discovered that modern packaging approaches do not support standalone scripts as the old approach did.

The code from 4 of the 5 scripts has been moved into modules within `mpas_tools` (the 5th script, the paraview VTK extractor, was already a stub).  They are inside of functions, as required for entry points into a Python package, in all cases.  The original scripts are preserved as stubs that call the entry points.

The entry points cannot have the same names as the original scripts: the `.py` extention is not allowed.

A few fixes have been made to update the coastal alteration and MOC modules for better PEP8 compliance.

The documentation has been updated to reflect the new names of the command-line tools.

### DIFF
--- a/conda_package/mpas_tools/ocean/coastline_alteration.py
+++ b/conda_package/mpas_tools/ocean/coastline_alteration.py
@@ -38,8 +38,13 @@ def main_add_critical_land_blockages():
     """
     Entry point for add_critical_land_blockages command-line tool
     """
+    description = """
+Add transects that identify critical regions where narrow strips of land block
+ocean flow.  These are, essentially, the opposite of critical passages, which
+must remain open for ocean flow.
+"""
     parser = \
-        argparse.ArgumentParser(description=__doc__,
+        argparse.ArgumentParser(description=description,
                                 formatter_class=argparse.RawTextHelpFormatter)
     parser.add_argument("-f", "--input_mask_file", dest="input_mask_filename",
                         help="Mask file that includes cell and edge masks.",
@@ -109,8 +114,13 @@ def main_widen_transect_edge_masks():
     """
     Entry point for widen_transect_edge_masks command-line tool
     """
+    description = """
+Alter transects to be at least two cells wide.  This is used for critical
+passages, to avoid sea ice blockage.  Specifically, mark cells on both sides
+of each transect edge mask as a water cell.
+"""
     parser = \
-        argparse.ArgumentParser(description=__doc__,
+        argparse.ArgumentParser(description=description,
                                 formatter_class=argparse.RawTextHelpFormatter)
     parser.add_argument("-f", "--mask_file", dest="mask_filename",
                         help="Mask file with cell and edge transect masks.",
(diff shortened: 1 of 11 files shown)


xylar:
Tested as part of https://github.com/MPAS-Dev/compass/pull/881.

xylar:
@sbrus89, can I ask you to represent the Ocean team by reviewing this (and https://github.com/MPAS-Dev/compass/pull/881 if you like)?

xylar:
You can get these changes by installing the 1.0.0rc2 package from the `conda-forge/label/mpas_tools_dev` channel.

sbrus89:
@xylar, sure I can take a look. Thanks for making these changes.

conda_package/mpas_tools/ocean/coastline_alteration.py
```diff
@@ -78,6 +110,44 @@ def widen_transect_edge_masks(dsMask, dsMesh, latitude_threshold=43.0):
...
+                        required=True)
+    parser.add_argument("-o", "--out_file", dest="out_filename",
+                        help="Output mask file,different from input filename.",
+                        metavar="MASKFILE",
+                        required=True)
+    parser.add_argument("-l", "--latitude_threshold",
+                        dest="latitude_threshold",
+                        help="Minimum latitude, degrees, for transect "
+                             "widening.",
+                        required=False, type=float, default=43.0)
```
sbrus89, line 140: Is there any reason for this default value?

sbrus89 (approved):
These changes look good to me. I'm not an expert on the conda changes, but I looked through the modifications to the scripts/entry points and approve by inspection.

conda_package/setup.py
```diff
@@ -71,12 +71,7 @@
                'landice/mesh_tools_li/create_landice_grid_from_generic_MPAS_grid.py',
                'landice/mesh_tools_li/define_cullMask.py',
                'landice/mesh_tools_li/interpolate_to_mpasli_grid.py',
```
sbrus89, line 73: Are changes needed for these other scripts as well?

conda_package/recipe/build.sh
```diff
@@ -3,13 +3,11 @@
...
 set -e
 
-cp -r ocean landice visualization mesh_tools conda_package
-
 cd conda_package
 ${PYTHON} -m pip install . --no-deps --no-build-isolation -vv
 
 # build and install ocean topography smoothing tool
-cd ${SRC_DIR}/conda_package/ocean/smooth_topo
+cd ${SRC_DIR}/ocean/smooth_topo
```
sbrus89, line 10: I don't understand the changes in this file and how they relate to the main objective of the PR, but I'm sure that's just because of my relative conda package ignorance.

conda_package/setup.py
xylar: Those are in #596, which I haven't asked you to review.

conda_package/recipe/build.sh
xylar: We were copying the `ocean` and several other directories into the `conda_package` directory so the tools found there could be included in the conda package.  A major purpose of this PR is to no longer make those copies but rather to require that the conda package takes no python code from outside the `conda_package` directory.  So that change is on purpose.

xylar: We can still build the C++ and Fortran tools that are outside of `conda_package` but we need to point to them where they are, not where they used to be copied to.

sbrus89: That makes total sense, thanks!

conda_package/setup.py
sbrus89: I see, thanks!

xylar:
Thanks @sbrus89! I really appreciate the prompt review!

PR #599, opened 2025-01-17
Use separate conda cache directories in CI

altheaden:
This PR updates the build workflow to use separate directories for each python version for conda caching. Issues with this seem to have been crashing CI runs lately.

I also fixed a bug in the docs workflow, where it was referring to a nonexistent python version matrix (likely a remnant of copying from the build workflow). Instead, I set this to be an environment variable called `PYTHON_VERSION` which will allow us to more easily specify and update the python version we use for the documentation without having to change it in a couple different places.


xylar (commented):
@altheaden, looks good!

PR #608, opened 2025-03-31
Bump cmake_minimum_required to 3.5 to make conda happy

xylar:
I'm not able to build the conda package at the moment because `cmake_minimum_required < 3.5` in various places.  I don't think there's a strong reason that we need this minimum so I'm proposing we bump it up to what conda wants.


xylar:
@mgduda, can you let me know if you've seen this, if you think you'd have time to review, and if so on what time scale?  That would help me know how soon I can make progress on other MPAS-Tools work that this is blocking.

xylar:
This seems to be a requirement of CMake >= 4.0, which was recently released. https://cmake.org/cmake/help/latest/release/4.0.html#id12

mgduda (approved):
Looks good to me!

xylar:
Thanks @matthewhoffman and @mgduda!

PR #609, opened 2025-04-02
Remove undefined variables that cause errors when calculating masks from thickness and bedTopography

trhille:
Remove undefined variable "run" in f-strings that causes and error when trying to calculate masks from thickness and bedTopography. Also add missing masks when calculating masks using thickness and bedTopography.


trhille:
Testing
-------
I was able to use this to make this figure, where `cellMask` is missing from the initial condition file (left-hand panel) and thus masks need to be calculated from thickness and bedTopography in order to plot the grounding line:
image

matthewhoffman (approved):
@trhille , thanks for taking care of this and thanks for providing the testing results.  Looks good to me.

PR #610, opened 2025-04-04
Switch setup configurations to `pyproject.toml`

altheaden:
In this PR I've switched the setup configurations from the `setup.py` approach to the more updated `pyproject.toml` approach that we're adopting in other repos.


xylar (approved):
@altheaden, this looks excellent!

PR #611, opened 2025-04-04
Require jigsaw and jigsawpy to be installed from source

xylar:
The conda-forge versions of `jigsaw` and `jigsawpy` are no longer being updated.  Instead, when we need them, we need to install them from source.  This PR adds a `build_jigsaw` command-line tool and function that can be used to do this.

It also removes `jigsaw` and `jigsawpy` as dependencies, and imports `jigsawpy` only when `jigsaw_driver()` is explicitly called, not at the top of the module.  This means that MPAS-Tools will typically not have `jigsaw` and `jigsawpy` installed, and downstream tools will need to explicitly decide how to handle these dependencies if they are required.

Documentation has been updated to reflect these changes.


xylar:
I tested both Compass and Polaris with this change.  They already install their own JIGSAW and JIGSAW-Python but they needed to be updated so they don't fail when there's not jigsaw/jigsawpy to uninstall.  With that fix, they work fine.

PR #614, opened 2025-04-05
Add class and command-line tool for converting from MPAS to XDMF + HDF5

xylar:
This is an alternative format to VTK for visualization in ParaView.  It offers several benefits:
* It is more disk efficient -- the mesh is stored only once, not once per time step
* It handles geometry of boundary vertices (mostly) correctly -- a bit more debugging is needed here
* It creates 2 files for each type of geometry (cells, edges and/or vertices) rather than also creating one file per time step

Several features of the ParaView VTK extractor have not yet been ported, including:
* culling periodic boundaries
* transforming to lat/lon space
* extracting topography geometry
* culling datasets to regions


xylar:
This worked fine in my testing with some relatively low resolution meshes (QU240 and Icos30).  I tested with cell, edge and vertex quantities.  I haven't done any substantial testing with time series yet.  But I think this can go in as the first draft.  No one is counting on this feature yet anyway.

PR #615, opened 2025-04-05
Preserve order of cells and edges on vertex in cull

xylar:
When culling cells, we wish to preserve the order of cells and edges on a vertex after culling, and just indicate which are invalid with zero indices.  This approach is more useful for visualization and in situations where the ordering between cells and edges around a vertex is important.


xylar:
I don't think this code is doing what I want.  More work is needed.

PR #620, opened 2025-04-11
Switch CI from miniconda to micromamba

altheaden:
This PR switches CI to use the `setup-micromamba` GitHub action instead of the `setup-miniconda` GitHub action, in an effort to fix CI crashes caused by the latest `miniconda` version.

### DIFF
--- a/.github/workflows/build_workflow.yml
+++ b/.github/workflows/build_workflow.yml
@@ -87,14 +87,16 @@ jobs:
 
       - if: ${{ steps.skip_check.outputs.should_skip != 'true' }}
         name: Create Build Environment
-        uses: conda-incubator/setup-miniconda@v3
+        uses: mamba-org/setup-micromamba@v2
         with:
-          activate-environment: "build"
-          miniforge-version: latest
-          channels: conda-forge
-          channel-priority: strict
-          auto-update-conda: true
-          python-version: 3.12
+          environment-name: mpas_tools_dev
+          init-shell: bash
+          condarc: |
+            channel_priority: strict
+            channels: 
+                - conda-forge
+          create-args: >-
+            python=${{ matrix.python-version }}
 
       - if: ${{ steps.skip_check.outputs.should_skip != 'true' }}
         name: Finalize Build Environment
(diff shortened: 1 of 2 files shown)


altheaden:
@xylar Considering the recent changes to remove `sphinx-multiversion`, is there anything else we should update in the docs workflow? I imagine we want to remove it from the list of packages to install during the environment setup at least.

xylar:
@altheaden, I think I did remove `sphinx-multiversion`:
https://github.com/MPAS-Dev/MPAS-Tools/pull/617/files
Did I miss somewhere?

xylar:
Oh, I see.  yes, it needs to be removed.

.github/workflows/build_workflow.yml
```diff
@@ -105,14 +107,13 @@ jobs:
       - if: ${{ steps.skip_check.outputs.should_skip != 'true' }}
         name: Install mpas_tools
         run: |
-          conda create -y -n mpas_tools_dev -c ${CONDA_PREFIX}/conda-bld/ \
-              -c conda-forge mpas_tools python=${{ matrix.python-version }} \
-              sphinx sphinx-multiversion mock sphinx_rtd_theme
+          conda install -y -c ${CONDA_PREFIX}/conda-bld/ \
+            mpas_tools python=${{ matrix.python-version }} \
+            sphinx sphinx-multiversion mock sphinx_rtd_theme
```
xylar, line 112: This should be updated to match:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/conda_package/dev-spec.txt#L37-L39
```suggestion
            sphinx mock sphinx_rtd_theme
```

altheaden:
@xylar I think I must have re-introduced it when I was rebasing onto your newest changes. I'll take care of it.

xylar:
No, I think I just missed it.

altheaden:
Ah, yeah, I see what happened now. There's a call to it at the very end of the workflow where we build the docs to check for issues. Should I just remove the whole job or replace that call with something else?
https://github.com/MPAS-Dev/MPAS-Tools/blob/964681c0eb3f81f7d8b2008d0b13f2899ed772e6/.github/workflows/build_workflow.yml#L114-L120

xylar:
Oops, I didn't realize which workflow this was.  That last docs build command needs to be the same as [in the documentation](https://mpas-dev.github.io/MPAS-Tools/master/building_docs.html):
```
cd conda_package/docs
DOCS_VERSION=master make versioned-html
```

xylar:
Or maybe we can just make `master` be `test` instead, since that seems more truthful.

altheaden:
@xylar How does this look?
https://github.com/MPAS-Dev/MPAS-Tools/blob/ee60eabc3598d22a139ec7c60689febb960e3ee0/.github/workflows/build_workflow.yml#L114-L118

xylar (approved):
Yes, exactly.  That looks good!

PR #621, opened 2025-04-13
Fix the barotropic streamfunction calculation

xylar:
The fix involves:
* detecting boundary edges
* finding all the unique loops for each continent or island
* finding a unique list of pairs (excluding the edge that
  would close the loop
* creating a sparse matrix with:
  * A Poisson equation for the BSF in terms of the vertically
    integrated vorticity at each vertex
  * Adding a constraint on each pair of boundary vertices found
    above ensuring that their values are equal
  * Adding a constraint that the BSF at the first vertext is
    zero
  * Adding the transpose of these boundary conditions so that
    the matrix is square
* the right-hand side involves the vertically integrated vorticity
  and zeros for the boundary conditions
* the solution involves the BSF and a Lagrange multiplier (dropped)
  for each boundary condition

### DIFF
--- a/conda_package/dev-spec.txt
+++ b/conda_package/dev-spec.txt
@@ -13,6 +13,7 @@ inpoly
 libnetcdf
 matplotlib-base>=3.9.0
 netcdf4
+networkx
 numpy>=2.0,<3.0
 progressbar2
 pyamg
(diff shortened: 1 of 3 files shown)


xylar:
I'm now seeing constant values along boundaries in ParaView in cases that didn't have them before the fix.  I'll post MPAS-Analysis plots once I have them.

xylar:
This worked as expected in testing in https://github.com/MPAS-Dev/MPAS-Analysis/pull/1074.

PR #632, opened 2025-05-27
Update `jigsaw_to_netcdf()` for efficiency and robustness 

xylar:
This merge:
* switches to xarray instead of netcdf4
* it does not hard-code `NETCDF3_CLASSIC` format but instead uses `write_netcdf()` with the default format.  This will be important for meshes with large numbers of vertices.
* it uses numpy arrays directly rather than looping over vertices

Some formatting has been cleaned up for better consistency with
other parts of `mpas_tools`.

To enable these changes, `circumcenter()` has also been updated to work with numpy arrays.

### DIFF
--- a/conda_package/mpas_tools/mesh/creation/jigsaw_to_netcdf.py
+++ b/conda_package/mpas_tools/mesh/creation/jigsaw_to_netcdf.py
@@ -1,13 +1,11 @@
-from __future__ import absolute_import, division, print_function, \
-    unicode_literals
+import argparse
 
 import numpy as np
+import xarray as xr
 
-from netCDF4 import Dataset as NetCDFFile
+from mpas_tools.io import write_netcdf
 from mpas_tools.mesh.creation.open_msh import readmsh
 from mpas_tools.mesh.creation.util import circumcenter
-
-import argparse
 
 
 def jigsaw_to_netcdf(msh_filename, output_name, on_sphere, sphere_radius=None):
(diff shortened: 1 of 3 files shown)


xylar:
@matthewhoffman and @trhille, I would like your review on this as it will affect your workflows.  I'm pretty confident in the work but we should test it on a more complicated mesh.  If you tell me which mesh(es) you'd like to have this tested on, I can run a before-and-after comparison in Compass to make sure the resulting meshes are identical.

xylar:
For the Antarctic mesh, differences are small:
```
  - xVertex: Data mismatch (max abs diff: 3.0081719160079956e-07)
  - yVertex: Data mismatch (max abs diff: 3.197201294824481e-07)
```
But I would have expected them to be exactly identical.  It seems I have some work to do.

xylar:
@matthewhoffman and @trhille, I don't seem to be able to get this to produce bit-for-bit identical results.  It seems like vectorizing things just inherently leads to order-of-operations changes.

I do think the results are identical to machine precision.  As an example, here are the vertices in the Antarctic mesh with the biggest change in `xVertex`:
```
Variable 'xVertex' differs. Max absolute difference: 3.0081719160079956e-07, at index: 466915
```
old method:
```
(2149013.33940856, 1443042.57600619, 0.0)
```
new method:
```
(2149013.33940826, 1443042.57600626, 0.0)
```

And similarly the biggest change in `yVertex`:
```
Variable 'yVertex' differs. Max absolute difference: 3.197201294824481e-07, at index: 794792
```
old method:
```
(2573113.05578246, -97387.13758518, 0.0)
```
new method:
```
(2573113.05578246, -97387.13758486, 0.0)
```

If these differences are big enough to bother you, I think there are strategies for vectorizing the old circumcenter approach that I could get working. It would be slower and more complicated than the approach I implemented here but this isn't a major bottleneck of our process at least for the size of mesh we currently have.  I'm mostly trying to anticipate problems down the road as I make use of this code in Polaris.

xylar:
You could always replace the vertex locations that correspond to cells that are unchanged with the values from the previous mesh in this scenario if it's important.  But I think it will be hard to get a mesh with identical cell locations out of JIGSAW anyway.

matthewhoffman:
@xylar , I agree with everything that @trhille said.  I have no concerns about these small differences breaking backwards reproducibility.

xylar:
Thanks @trhille and @matthewhoffman!!

PR #633, opened 2025-05-27
Add support for `NETCDF3_64BIT_DATA` in `write_netcdf()`

xylar:
For performance, this involves first writing a `NETCDF4` file and then using `ncks` to convert it to `NETCDF3_64BIT_DATA` (CDF5) format.

This merge also:
* adds a conversion of `int64` format to `int32`, since the MPAS framework doesn't support `int64`
* adds documentation for `write_netcdf()`
* adds unit tests for `write_netcdf()`, including both CDF5 and int64 --> int32 capabilities
* adds `nco` as a dependency
* adds the logger to `write_netcdf()` calls in the mask creator (to keep output from conversion to CDF5 when it happens)

### DIFF
--- a/conda_package/mpas_tools/io.py
+++ b/conda_package/mpas_tools/io.py
@@ -1,11 +1,12 @@
-from __future__ import absolute_import, division, print_function, \
-    unicode_literals
-
-import numpy
-import netCDF4
-from datetime import datetime
+import os
+import subprocess
 import sys
+from datetime import datetime
 
+import netCDF4
+import numpy
+
+from mpas_tools.logging import check_call
 
 default_format = 'NETCDF3_64BIT'
 default_engine = None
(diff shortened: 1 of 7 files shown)


xylar:
## Testing in Polaris

I was able to run the `pr` test suite on Chrysalis with these changes and by branch https://github.com/xylar/polaris/tree/write-cdf5, showing that the results were bit-for-bit identical to those with the Polaris `main` branch without these changes, and also that files were written in CDF5 format, as expected.

xylar:
## Further Testing

I have further tested these changes in a full workflow in Polaris that involved culling topography.  This exposed issues with `string` variables in the python mask creator tools that I have now fixed as part of this PR.

conda_package/mpas_tools/io.py
```diff
@@ -88,14 +117,50 @@ def write_netcdf(ds, fileName, fillValues=None, format=None, engine=None,
...
+        ]
+        if logger is None:
+            subprocess.run(
+                args,
+                check=True,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+            )
+        else:
+            check_call(args, logger=logger)
```
xylar, line 153: Then, delete the intermediate file!

PR #644, opened 2025-06-14
Improve performance of barotropic streamfunction 

xylar:
This merge refactors the barotropic streamfunction, separating out computation of vertically integrated velocity and vorticity into their own modules.  It improves performance and memory usage for large meshes by:
* chunking along the `nEdges` dimension during computation of the vertically integrated velocity
* saving to intermediate files if a temporary directory is provided
* loading numpy array for subsequent (2D) computations

### DIFF
--- a/conda_package/docs/api.rst
+++ b/conda_package/docs/api.rst
@@ -227,6 +227,7 @@ Ocean Tools
    depth.compute_zmid
 
    compute_barotropic_streamfunction
+   shift_barotropic_streamfunction
 
 .. currentmodule:: mpas_tools.ocean.inject_bathymetry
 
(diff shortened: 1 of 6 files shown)


xylar:
## Testing

I was able to compute the Barotropic streamfunciton on the `RRSwISC6to18E3r5` mesh with this branch, whereas MPAS-Analysis ran out of memory in runs without this fix.

https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.xylar/analysis_testing/BSF/20250501.wcycl1850.ne120pg2_r025_RRSwISC6to18E3r5.test5.chrysalis/

PR #656, opened 2025-08-05
Fix `pre-commit` workflow and other CI updates/cleanup

altheaden:
This PR makes a few changes and updates to CI.
1. `PYTHON_VERSION` is now defined in all workflow files for consistency
2. `REVIEWERS` variable is now used to list the reviewers instead of a magic string to make things more readable
3. `pre_commit_update_workflow.yml` has been fixed to use conda in to avoid a bug with `gh` which was causing permissions errors in workflow runs
4. `workflow_dispatch` has been added to `pre_commit_update_workflow.yml` CI workflow to allow manual triggering
5. `pre_commit_update_workflow.yml` now checks for and closes any existing update PRs before creating a new one, something that previously would have caused a failure
6. @andrewdnolan has been added to the dependabot reviewers list

@xylar Review checklist:
* [ ] Add @andrewdnolan as maintainer (if he isn't already)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,4 +11,5 @@ updates:
     reviewers:
       - "xylar"
       - "altheaden"
+      - "andrewdnolan"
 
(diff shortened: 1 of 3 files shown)


xylar (approved):
Looks great!  Thanks @altheaden!

PR #662, opened 2025-08-15
Remove encoding unlimited_dims if Time not present

xylar:
When xarray 2025.08.0 (released recently) opens a file with an empty unlimited Time dimension but no variables with Time, it drops `Time` as a dimension but keeps it in `unlimited_dims`.  This is probably not an expected behavior but it's easy enough to address by setting `unlimited_dims = None` if `Time` is not present.

This merge also switches `sort_mesh` to use `write_netcdf()` to avoid this issue.  A `--format` argument has been added that defaults to `NETCDF4`, which was previously the hard-coded format.


xylar:
This issue was caught by CI, so having CI passing will be the the test to make sure this is fixed.  But I will also run Polaris workflows with a test build of mpas_tools 1.3.0 before making the release.

PR #663, opened 2025-08-15
Use environment path not name in `build_jigsaw`

xylar:
This pull request makes a targeted change to the environment activation logic in the `build_jigsaw` function. The update replaces the use of the `CONDA_DEFAULT_ENV` environment variable with `conda_env_path`, streamlining how the conda environment is activated and removing the associated error handling.


xylar:
It looks like this fix is needed for OSX builds of jigsaw to work.  Testing in:
https://github.com/conda-forge/mpas_tools-feedstock/pull/131

PR #675, opened 2025-09-05
Add script to convert cell coordinates from plane to sphere

matthewhoffman:
Script to allow plotting MALI output on a sphere in Paraview. This works by replacing x/y/zCell fields and modifying global attributes so Paraview will plot as a spherical mesh.  Paraview only uses those 3 fields for plotting.  Plotting MALI output on a sphere allows it to be visualized with output from other E3SM components (e.g. MPAS-Ocean).

I placed it in landice/mesh_tools_li because it modifies a mesh, even though we are likely to primarily use it on output files.


matthewhoffman:
Example of plotting ice speed from the MALI 4km AIS mesh on a sphere overlaid on bottom depth from MPAS-Ocean:
image

matthewhoffman:
Using optional surface warping:

image

landice/mesh_tools_li/convert_cell_coordinates_to_sphere.py
```diff
@@ -0,0 +1,78 @@
...
+
+if args.warp_surface:
+    if 'upperSurface' in ds.variables:
+        sfc = ds.variables['upperSurface'][0,:]
+    else:
+        thickness = ds.variables['thickness'][0,:]
+        bedTopography = ds.variables['bedTopography'][0,:]
+        sfc = np.maximum(bedTopography + thickness,
+                         (1.0 - 910.0 / 1028.0) * thickness)
+        #sfc = thickness + np.minimum(bedTopography, 0.0) * 1028.0 / 910.0
```
trhille: remove

trhille (changes requested):
Thanks for this very useful script, @matthewhoffman. I think we want to avoid editing files in-place because we can imagine a nightmare scenario in which someone inadvertently overwrites one of the mesh files in our database, so almost all of my comments have to do with that.

trhille (approved):
Thanks for the revisions, and apologies for the slowness on my end.

PR #710, opened 2026-04-01
 Fix computation of `angleEdge` in NetCDF-C mesh converter 

xylar:
The old spherical `angleEdge` calculation in `MpasMeshConverter.x` inferred the
edge angle from the latitude difference between edge vertices and then used a
sign correction based on a nearby artificial "north pole" point. This
approximation becomes inaccurate and unstable near the poles, where local
east/north directions are poorly represented by latitude differences.

This merge replaces that logic with a Cartesian geometric formulation:
compute the edge normal from the vertex chord and edge position on the
sphere, project it onto the local east/north basis at the edge, and
compute `angleEdge` with `atan2`.  This matches the downstream Polaris
fix and gives robust spherical angles across the globe, including polar
regions.

This merge also adds `mpas_tools.mesh.spherical` with utilities to compute local east/north
vectors, edge-normal vectors, and recomputed spherical `angleEdge` values from
mesh geometry. These routines mirror the corrected spherical formulation used
for the C++ mesh converter and provide a reusable Python reference
implementation.

Finally, this merge adds a regression test to make sure a QU1920 mesh gets the same `angleEdge` from the python and C++ codes.

### DIFF
--- a/conda_package/docs/api.rst
+++ b/conda_package/docs/api.rst
@@ -76,6 +76,15 @@ Mesh conversion
    compute_mpas_flood_fill_mask
    compute_lon_lat_region_masks
 
+.. currentmodule:: mpas_tools.mesh.spherical
+
+.. autosummary::
+   :toctree: generated/
+
+   recompute_angle_edge
+   calc_edge_normal_vector
+   calc_vector_east_north
+
 .. currentmodule:: mpas_tools.merge_grids
 
 .. autosummary::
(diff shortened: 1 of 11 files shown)


cbegeman (approved):
Approving based on visual inspection. Did you test any meshes with finer resolutions than 1290km?

xylar:
@cbegeman, I tested this in a similar way (C++ vs. python) on an EC30to60 mesh, again without any problems.

I also added some fixes to the documentation I noticed when I was trying to install the conda package into my local pixi environment.

PR #717, opened 2026-04-28
Change default earth_radius_km value for the hex_projection tool to 6371.229 km

mgduda:
This PR changes the default `earth_radius_km` value for the `hex_projection` tool to 6371.229 km.

MPAS-Atmosphere uses a sphere radius of 6371229.0 m, and since the `hex_projection` tool is intended primarily for use with MPAS-Atmosphere, this PR sets the default sphere radius provided in the `namelist.projections` file for the `hex_projection` tool to match that value.


skamaroc (approved):
Agreed - we should be consistent with our current radius in mpas

abishekg7 (approved):

PR #719, opened 2026-05-08
Improve performance in python masking

xylar:
### **1\. Vectorized geometry creation (\~17× speedup on polygon building)**

**\_get\_polygons**: replaced the Python for loop building `shapely.geometry.Polygon` objects one-by-one with a single `shapely.polygons(coords_3d)` call. For a 1 M-cell mesh this drops from \~11.5 s to \~0.68 s.

**Three region/projection functions**: same pattern for points — shapely.points(numpy.stack(\[lon, lat\], axis=-1)) replaces the old list comprehension.

### **2\. Build STRtree once, query per feature (\~20× speedup for 30 transects)**

**\_compute\_transect\_masks** and **\_compute\_region\_masks**: the old code (through \_compute\_mask\_from\_shapes) rebuilt an STRtree from a *small chunk* of polygons for every chunk × every feature. A 500 K-cell mesh with 30 transects now builds one STRtree (0.12 s) and queries it 30 times (0.011 s total) instead of rebuilding \~15 000 small trees.

### **3\. Thread-parallel feature queries (no serialization overhead)**

When pool is provided, a ThreadPoolExecutor with the same worker count is used to run feature queries in parallel. Shapely 2.x releases the GIL during geometric operations, so threads genuinely run concurrently without the pickle/IPC overhead that made the old multiprocessing chunk approach expensive.

### **4\. Flood-fill via scipy.connected\_components (\~16× speedup)**

**\_flood\_fill\_mask**: the Python while True BFS loop (O(diameter × cells) Python iterations) is replaced by a pure-C scipy sparse graph path: build the adjacency in numpy → connected\_components → keep components that contain a seed cell. Tested at 0.007 s vs 0.113 s on a 40 000-cell grid.

### **5\. Removed dead code**

\_compute\_mask\_from\_shapes, \_contains, and \_intersects are deleted; functools.partial and progressbar imports removed with them.

### DIFF
--- a/conda_package/mpas_tools/mesh/mask.py
+++ b/conda_package/mpas_tools/mesh/mask.py
@@ -21,14 +21,16 @@ See function docstrings for details.
 """
 
 import argparse
-from functools import partial
+from concurrent.futures import ThreadPoolExecutor
 
 import numpy
-import progressbar
+import shapely
 import shapely.geometry
 import xarray as xr
 from geometric_features import read_feature_collection
 from igraph import Graph
+from scipy.sparse import csr_matrix
+from scipy.sparse.csgraph import connected_components
 from scipy.spatial import KDTree
 from shapely.geometry import GeometryCollection, MultiPolygon, Polygon, box
 from shapely.strtree import STRtree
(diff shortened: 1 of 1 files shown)


xylar:
Mask creation (particularly for transects) has been a bottleneck in culling MPAS meshes in Compass and Polaris.  My hope with this work is that we can reduce that bottleneck, perhaps significantly for large meshes.

xylar:
I ran the `e3sm/init/icos30km/topo/cull` test from Polaris on Chrysalis.  Without these changes, I see:
```
  * step: cull_mask
          execution:        SUCCESS
          runtime:          0:14:33
```
With them, I see:
```
  * step: cull_mask
          execution:        SUCCESS
          runtime:          0:00:48
```
That's astounding!  Results are bit-for-bit.

PR #721, opened 2026-05-12
Include density correction in SL axis in globalStats

matthewhoffman:
This merge adds the density correction for ice to melt into freshwater rather than seawater when plotting the secondary y-axis for sea-level equivalent in the globalStats plotting script.  It also updates to the Gregory et al. (2019) recommended ocean area value.  The first item decreases scaling of the secondary axis by about 3%.  The second increases it by about 1%.  These changes only affect the secondary axis scaling and do not affect the actual data.  Note that a full incorporation of the density change of melted ice (i.e. ice below flotation and fully floating ice) would need to occur in MALI and would not support a 1:1 relationship between Gt of VAF lost and sea-level equivalent.  This secondary axis will always be approximate for that reason.

### DIFF
--- a/landice/output_processing_li/plot_globalStats.py
+++ b/landice/output_processing_li/plot_globalStats.py
@@ -14,8 +14,10 @@ import matplotlib.pyplot as plt
 import textwrap
 
 rhoi = 910.0
-rhosw = 1028.
-rhofw = 1000.
+rhosw = 1028.0
+rhofw = 1000.0
+Aocn = 3.625e14  # m^2 (Gregory et al. 2019)
+m_to_mm = 1000.0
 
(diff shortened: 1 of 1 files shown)


matthewhoffman:
# Testing

I created a globalStats file sets two time levels for `volumeAboveFloatation`: one at the m^3 equivalent to 362.5 Gt (398351648352 m3) and at 0.  When plotting change in volume above floatation with this script, this should show a change of exactly 1 mm.  Before this PR, the result was slightly off:
Screenshot 2026-05-12 at 2 13 44 PM
With this PR it lines up:
Screenshot 2026-05-12 at 2 14 08 PM

trhille (approved):
Thanks, @matthewhoffman! Approving based on inspection and your testing.

PR #726, opened 2026-05-28
Add script to interpolate ITS_LIVE dh/dt and surface fields to MALI mesh

trhille:
Add script to interpolate ITS_LIVE dh/dt and surface fields to MALI mesh for a user-specified time period. Based on the ITS_LIVE data set connected with https://doi.org/10.5194/essd-18-1729-2026. Note that this does not automatically update the `thickness` field based on the average surface field calculated from the ITS_LIVE data, but that can easily be done with an `ncap2` command if desired..


trhille:
Testing
-------
Testing on a 4–40km Greenland mesh:

image
image
image
image


This runs in just a few minutes on one PM compute node.

PR #732, opened 2026-06-02
Reuse extrapolation weights for multiple variables

trhille:
This merge calculates IDW weights during the first extrapolation of a variable across the MPAS domain when converting from a Albany exodus file and then reuses those weights for the extrapolation of subsequent variables. This eliminates the repeated wave-front search that can be slow on large meshes, replacing an effectively O(N^1.5) operation with O(N) operation.

### DIFF
--- a/landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
+++ b/landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
@@ -145,6 +145,7 @@ vars_not_converted = []
 
 # Loop through the variables, convert from Albany to MPAS, interpolate temperature
 # from Albany to MPAS vertical layers, extrapolate and smooth beta and stiffnessFactor
+extrapolation_pattern = None  # Cache for extrapolation pattern (cell order + neighbor info)
 for var_name in var_names:
     #set appropriate methods for smoothing and extrapolation
     if var_name == "beta":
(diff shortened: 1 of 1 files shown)


trhille:
Using these changes on a fairly small 4–40km GIS mesh (nCells = 55928), I get:
```
Time taken for extrapolation and smoothing muFriction: 15.169 seconds
Time taken for extrapolation and smoothing stiffnessFactor: 0.974 seconds
```
Without these changes:
```
Time taken for extrapolation and smoothing muFriction: 8.124 seconds
Time taken for extrapolation and smoothing stiffnessFactor: 14.412 seconds
```
(I ran this three times for each configuration. Times varied within about ±1.5s compared with those reported above.)
So, that's a large fractional savings, but a small absolute savings. However, as we go to larger meshes, this savings should become considerable. Here is Claude's estimate:


"The key is the algorithmic complexity of each approach:

Original (per variable): O(N × R), where R is the number of wave-front rings needed to fill all empty cells. For a 2D mesh, R ≈ √N (the "radius" of the extrapolation region in cells). So effectively O(N^1.5) — each of the 24 iterations scans all remaining empty cells (O(N)), and there are O(√N) such iterations.

Cached replay: A single ordered pass over all filled cells — O(N).

So the savings ratio grows as O(√N). Concretely:

image

For typical large Antarctic meshes (hundreds of thousands to millions of cells), the cache turns a multi-hour extrapolation into minutes for the second and third variables. The first variable still pays full cost — which itself grows as O(N^1.5) — so on very large meshes, the pattern-building step will also become a bottleneck."

trhille:
Testing
-------
These changes yields BFB behavior with the current behavior in `master` for variables with `min` extrapolation (i.e., muFriction):
image

There are roundoff-level differences with the stiffnessFactor field, which uses IDW extrapolation, but these can be ignored:
image

PR #733, opened 2026-06-22
Add ISMIP7 postprocessing framework

hollyhan:
This PR updates the MALI postprocessing scripts for ISMIP7 based on the ISMIP6 postprocessing scripts.

- Updated 2D flux processing to use MALI's online time-averaged flux outputs directly, removing the legacy offline time-averaging workflow.
- Updated mandatory 1D scalar flux outputs to match the ISMIP7 variable request:
  - Added `tendlibmassbfgr`
  - Updated `tendlifmassbf` to represent ice-front melting only
  - Removed the legacy total basal mass balance scalar output
- drop requirement for a reference ISMIP grid as an input and instead generate it by the script following definition provided by ISMIP
- refactor many repeated operations into shared functions
- support multifile datasets instead of requiring that the MALI output files are concatenated together prior to running this tool
- added support for GIS in addition to AIS
- added support for Arete in addition to DOE as institutions
- substantial adjustments to time handling to match ISMIP conventions
- general cleanup and reorganization and some low-hanging performance improvements (e.g. only including required variables in temp files to keep i/o, file size, and remapping cost down)
- quality of life improvements, e.g., progress indicator for stdout, writing output to a exp subdir so the destination path doesn't need to be changed for each experiment
- tested with MALI initial submission results for both AIS and GIS and run through ISMIP validation checker.

### Possible future work
- Implement the remaining optional ISMIP7 variables. (That is, variables highlighted in purple and `litempavg` in the ISMIP7 variable spreadsheet [https://docs.google.com/spreadsheets/d/1yWDxk8pRwwp3gvRT6j0lsWA_5hAkAXqZdLGtgpnLSNE/edit?gid=2113789754#gid=2113789754](url) 
- Fix calculation of the grounding line flux once it is clarified by ISMIP what the proper calculation is


matthewhoffman (approved):
I'm approving & merging because this has successfully been used to pass the compliance checker for the initial submission.  Further tweaks can be made in future PRs.

PR #736, opened 2026-06-26
Add support for Omega split mesh and vertical coordinate datasets

### LINKED ISSUE #735: Updates needed after separation of horiz mesh and vertical coord datasets for Omega
`compute_barotropic_streamfunction` assumes `maxLevelCell` is in the horizontal mesh dataset https://github.com/E3SM-Project/polaris/pull/626. There may be other functions that make similar assumptions.

xylar:
## Adding support for Omega

[Omega](https://github.com/E3SM-Project/Omega) is the next-generation ocean model being developed alongside MPAS-Ocean. A key difference is how Omega organizes its data: instead of bundling the horizontal mesh, the vertical coordinate, and the model state all into a single file (as MPAS-Ocean has historically done), Omega keeps these as **separate datasets**. This PR updates the ocean tools in MPAS-Tools to work with that split layout, so the same analysis and visualization tools can be used for both MPAS-Ocean and Omega output.

### What changed for users

- **A new, optional vertical-coordinate dataset.** The barotropic streamfunction tools (`compute_barotropic_streamfunction`, `compute_vertically_integrated_velocity`) and the transect plotting tool   (`plot_feature_transects`) now accept an optional `ds_vert_coord` argument for the vertical coordinate variables (`minLevelCell`, `maxLevelCell`, `bottomDepth`). If you don't provide it, it defaults to your mesh dataset, so **existing MPAS-Ocean workflows keep working unchanged**.

- **A new command-line option.** The `plot_feature_transects` command gains a `--vert_coord` option to point at a separate vertical-coordinate file.

### Backwards-incompatible changes (why this is version 2.0.0)

- **Layer thickness now comes from the data file, not the mesh file.** Layer thickness changes over time, so it is part of the model *state*, not the static mesh. Tools that previously read `layerThickness` from a mesh dataset now read it from the data (state) dataset. In the common case where your mesh and data are the same file, nothing changes. If you relied on passing a separate mesh file that carried `layerThickness`, that variable now needs to be in your data file. This also affects `add_zmid`, which now reads `layerThickness` from its input file rather than the coordinate file.

- **The 1D depth coordinate is no longer supported for Omega.** `add_depth` (and the underlying `compute_depth`) build an approximate 1D `depth` coordinate from `refBottomDepth`, which only makes sense for a z-level/z-star vertical coordinate. Omega uses general vertical coordinates and does not provide `refBottomDepth`, so these tools now raise a clear error explaining the situation and pointing you to the 3D `zMid` coordinate (`add_zmid`) instead. MPAS-Ocean z-star runs are unaffected.

### Testing

A pre-release, `2.0.0rc1`, has been tagged for downstream testing. Existing depth unit tests were updated to match the new layer-thickness behavior, and a new test confirms the clear error when `refBottomDepth` is missing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Fixes #735

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -80,4 +80,4 @@ authors:
 repository-code: 'https://github.com/MPAS-Dev/MPAS-Tools'
 url: 'https://mpas-dev.github.io/MPAS-Tools/master/'
 version: 2.0.0
-date-released: '2026-06-26'
+date-released: '2026-07-01'
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -79,5 +79,5 @@ authors:
     family-names: Smith
 repository-code: 'https://github.com/MPAS-Dev/MPAS-Tools'
 url: 'https://mpas-dev.github.io/MPAS-Tools/master/'
-version: 1.6.0
-date-released: '2026-05-09'
+version: 2.0.0
+date-released: '2026-06-26'
--- a/conda_package/docs/ocean/depth.rst
+++ b/conda_package/docs/ocean/depth.rst
@@ -17,6 +17,14 @@ directly if one has a suitable ``refBottomDepth`` data array indicating the
 reference depth of the bottom of each layer in a 1D coordinate that is
 independent of both time and horizontal coordinate.
 
+.. warning::
+
+    This 1D ``depth`` coordinate relies on ``refBottomDepth`` and therefore
+    assumes a z-level/z-star vertical coordinate.  It is **not supported for
+    Omega or other general vertical coordinates**, which do not provide
+    ``refBottomDepth``.  In those cases, use the 3D ``zMid`` coordinate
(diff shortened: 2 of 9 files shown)


xylar:
I will try to test this in Polaris soon, maybe over the weekend if I have some time to kill.

CITATION.cff
```diff
@@ -79,5 +79,5 @@ authors:
     family-names: Smith
 repository-code: 'https://github.com/MPAS-Dev/MPAS-Tools'
 url: 'https://mpas-dev.github.io/MPAS-Tools/master/'
-version: 1.6.0
-date-released: '2026-05-09'
+version: 2.0.0
+date-released: '2026-06-26'
```
xylar:
```suggestion
date-released: '2026-07-01'
```

xylar:
Tested in https://github.com/E3SM-Project/polaris/pull/632 as part of 2.0.0rc2.

PR #737, opened 2026-06-27
Add `open_dataset` and `open_mfdataset` wrappers to `mpas_tools.io`

xylar:
This PR adds thin wrappers around `xarray.open_dataset()` and `xarray.open_mfdataset()` that select the NetCDF engine from the module-level `mpas_tools.io.default_engine` variable when no engine is passed explicitly.

`xarray.open_dataset()` tries to open a file looking for "magic bits" to auto-select a backend, and that probe can crash on `NETCDF3_64BIT_DATA` (`CDF5`) files (see https://github.com/E3SM-Project/polaris/issues/624). Specifying an engine explicitly avoids the sniffing. Since xarray has no global default-engine setting, reusing the existing `default_engine` variable (already consumed by `write_netcdf()`) gives downstream tools a single, process-wide knob for both reading and writing without modifying every call site.

The logger argument is included for API symmetry with write_netcdf and future diagnostics; no error recovery is performed because the CDF5 failure is a hard crash that cannot be caught.

### DIFF
--- a/conda_package/docs/api.rst
+++ b/conda_package/docs/api.rst
@@ -151,6 +151,8 @@ I/O
    :toctree: generated/
 
    write_netcdf
+   open_dataset
+   open_mfdataset
 
 Parallelism
 -----------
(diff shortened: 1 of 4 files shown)


cbegeman:
@xylar Thanks!

xylar:
Tested (via the 2.0.0rc2 tag) in https://github.com/E3SM-Project/polaris/pull/632.

PR #742, opened 2026-07-23
Add functionality to MALI ocean-data interpolation script

stephenprice:
Minor changes to allow for:

1. Allow more general specification for ocean output file year (e.g. zero-padded, like year '0059')
2. Co-pilot generated changes to ensure that the script works to generate monthly outputs of interpolated values (previous functionality worked to interpolate annual averages from monthly mpas-o outputs; new functionality also allows for the creation of monthly resolved, interpolated outputs)

### DIFF
--- a/landice/mesh_tools_li/interpolate_mpasOcean_to_mali.py
+++ b/landice/mesh_tools_li/interpolate_mpasOcean_to_mali.py
@@ -146,7 +146,7 @@ class mpasToMaliInterp:
 
         # open and concatenate diagnostic dataset
         self.DS = xr.open_mfdataset(os.path.join(self.options.mpasoHistDir,
-            f'*.mpaso.hist.am.timeSeriesStatsMonthly.{year}-*-01.nc'), combine='nested', concat_dim='Time', decode_timedelta=False)
+            f'*.mpaso.hist.am.timeSeriesStatsMonthly.{year:04d}-*-01.nc'), combine='nested', concat_dim='Time', decode_timedelta=False)
         
         # variables for time averaging
         self.temperature = self.DS['timeMonthly_avg_activeTracers_temperature']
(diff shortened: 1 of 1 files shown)


landice/mesh_tools_li/interpolate_mpasOcean_to_mali.py
```diff
@@ -157,10 +157,8 @@ def get_data(self, year):
         avg_layerThickness = self.DS['timeMonthly_avg_layerThickness']
         self.layerThickness = avg_layerThickness.data
 
-        #xt = DS['xtime_startMonthly']
-        
-        #xtime = np.array([xt],dtype=('S',64)) 
-        # << NOTE >>: may need to eventually use: "xtime.data.tobytes().decode()" but not sure yet
+        # monthly xtime strings (one per month), used when not time averaging
+        self.xtimeMonthly = self.DS['xtime_startMonthly'].values
```
matthewhoffman, line 161: Is `xtime_startMonthly` in all monthly average MPAS-O history files?  I don't recognize the name, but maybe it is automatically generated by the time-averager in MPAS-O.  Did you do any preprocessing of the MPAS-O history files or did your run of this script work successfully with the unmodified history files?

stephenprice: Good question. 

``` 
char xtime_startMonthly(Time, StrLen) ;
char xtime_endMonthly(Time, StrLen) ;
```
... are in all `timeSeriesStatsMonthly` mpas-o output files (for the Max, Min files their names are altered accordingly).

But I did NO modification of these files prior to running this script. I appears to have worked on them as is.

Note that I have spot checked the outputs from this using ncdump and they look reasonable in the sense that there are a lot of zeroes and (fewer sets of) non-zeroes for the field `ismip6shelfMelt_3dThermalForcing`, which look like physically reasonable ocean temperature values. But we can let this PR sit until we test them to make sure. I just don't know enough about what these fields should look like or how to check them further

PR #748, opened 2026-08-20
Reconstruct vectors correctly on planar meshes 

xylar:
`reconstruct_variable()` computed the Cartesian (X, Y, Z) components correctly on any mesh but then applied the spherical rotation to get the zonal and meridional components unconditionally, using `latCell` and `lonCell`.  Planar meshes (for example from `mpas_tools.planar_hex`) have `latCell = lonCell = 0`, so the rotation collapsed to `Zonal = Y` and `Meridional = Z`, where Z is essentially zero.  Planar zonal fields were really the y component and planar meridional fields were noise.

Branch on the `on_a_sphere` attribute of the mesh instead, and on a planar mesh take `Zonal = X` and `Meridional = Y`, which is what MPAS' own `mpas_reconstruct_2d` and `mpas_reconstruct_1d` do.  Reading `latCell` and `lonCell` now happens only in the spherical branch, so a planar mesh that lacks them works too.  A mesh with no `on_a_sphere` attribute is treated as spherical, preserving the previous behavior.

### DIFF
--- a/conda_package/docs/vector.rst
+++ b/conda_package/docs/vector.rst
@@ -28,6 +28,14 @@ The command-line tool ``vector_reconstruct`` and the function
 reconstruct Cartesian (X, Y, Z), zonal and meridional components of an MPAS
 vector field at cell centers, given the field on edge normals.
 
+On spherical meshes, the zonal and meridional components are found by rotating
+the Cartesian components into the local geographic frame at each cell center,
+using ``latCell`` and ``lonCell``.  On planar meshes (indicated by the
+``on_a_sphere`` attribute of the mesh being ``NO``), no rotation is performed:
+the zonal and meridional components are the x and y components, respectively.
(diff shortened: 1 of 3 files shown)


xylar:
@mwarusz, could you have a look?  I can't make you an official reviewer but that's just a formality.

xylar:
Just worry about 30220e63e64cfd61335a7d8eb636a55d4d5e7bdf.  The other commit is just formatting.

xylar:
## Testing

I ran the Polaris Ekman column test using this fix and https://github.com/E3SM-Project/polaris/pull/715.

Before this fix, I was seeing:

image

Now, I'm seeing:

Image

By eye, this looks identical to the MPAS-Ocean solution now (which uses its own internal vector reconstruction):

Image

xylar:
@mwarusz, I'm going to go ahead with this fix.  We can revisit if need be.

PR #753, opened 2026-09-01
Vectorize MALI mesh flood-fill extrapolation for large-mesh speedup

trhille:
The per-cell Python neighbor loop used to extrapolate beta/muFriction/ stiffnessFactor into undefined regions is O(rings x remaining-empty-cells) and slow on large meshes. Replace it with a shared extrapolate_into_mask() helper that processes only the active BFS frontier each ring using vectorized numpy neighbor reductions, preserving the same min/idw/value semantics. Also vectorize the one-time grounded-neighbor dilation in extrapolate_variable.py.

On an 829k-cell GIS mesh this is ~107x faster for min (68.2s -> 0.64s, bitwise identical) and ~61x for idw (81.2s -> 1.32s, identical to 4e-15 sum-order roundoff). The new code also terminates cleanly on voids disconnected from any valid data, which the original loop could not.

### DIFF
--- a/landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
+++ b/landice/mesh_tools_li/conversion_exodus_init_to_mpasli_mesh.py
@@ -25,6 +25,48 @@ import os
 import sys
 from datetime import datetime
 
+
+def extrapolate_into_mask(varValue, keepCellMask, cellNeighbors,
+                          neighborValid, xCell, yCell, method):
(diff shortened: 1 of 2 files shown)


trhille:
Testing
-------
I tested this with muFriction on a 1to10km Greenland mesh using `min`, which reduced the total script run time from 77s to 14s. The results are bit-for-bit identical.

matthewhoffman (approved):
Sounds great!

PR #755, opened 2026-09-02
Add tool to remove disconnected ice clusters from a MALI mesh

trhille:
remove_disconnected_ice.py groups ice cells (thickness above a threshold) into connected components over the mesh cellsOnCell adjacency using scipy.sparse.csgraph.connected_components. A flood-fill seeded at a user-defined location (default: mesh centre; snaps to nearest ice cell if ice-free) identifies the main ice body, which is always preserved.

Disconnected clusters are zeroed out when they are below a size threshold (default 10 cells) OR when they are entirely floating (no grounded cells per the flotation criterion), so detached icebergs / ice-shelf fragments are removed regardless of size. Larger grounded clusters are left unchanged. Edits in place or writes to --outFile.

### DIFF
--- a/landice/mesh_tools_li/remove_disconnected_ice.py
+++ b/landice/mesh_tools_li/remove_disconnected_ice.py
@@ -1,0 +1,193 @@
+#!/usr/bin/env python
+'''
+Remove small, dynamically disconnected clusters of ice from a MALI mesh.
+
+Ice cells (thickness greater than a threshold) are grouped into connected
+components using the mesh cell adjacency (cellsOnCell).  A flood-fill is
+seeded at a user-defined location (default: the centre of the mesh) to
+identify the main ice body; the component containing the seed cell is always
+preserved.  Every other ice cluster that is not connected to the main body is
+considered disconnected.  Any disconnected cluster whose size is below a
+user-defined threshold (default: 10 cells) has its ice thickness set to zero.
+In addition, disconnected clusters that are entirely floating are removed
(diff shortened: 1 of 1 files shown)


trhille:
Testing
-------
I tested this on on a 1–10 km Greenland mesh that had many little disconnected lumps of ice that caused issues with the Albany optimization.
Mesh used in the optimization is on the left, with thickness = 0.01 m contoured in white and bedTopography=0.0 in black. Cyan dots show spd > 20km/yr. One the right is the mesh after running this script: `python /global/cfs/cdirs/fanssie/users/trhille/MPAS-Tools/landice/mesh_tools_li/remove_disconnected_ice.py -f /pscratch/sd/t/trhille/ISMIP7/GIS/mesh_gen_20260811/1to10km/optimization/mesh/GIS_1to10km_r03_20260828.nc -o GIS_1to10km_r03_20260902.nc -n 10`

```
Seed cell 609478 at (114481.3, -2002453.0); main ice body has 370756 cells
Found 122 disconnected ice cluster(s); removing 112 (floating: 1, below 10-cell threshold: 112)
Zeroed ice thickness on 212 cell(s) in 112 disconnected cluster(s)
Wrote 'GIS_1to10km_r03_20260902.nc'
```

image
This mesh contains 829112 cells and the script took about 50 seconds to run on a login node.

matthewhoffman (approved):
This looks good to me!  Copilot review came up with a few nitpicky things, but I dismissed them.

